Issue #8, opened 2023-10-27
bug: Marker counter after undo action doesn't reset when called again


noornee:
### Problem
I encountered an issue while using satty. 
If a user places a marker in the wrong spot and wants to remove it from that spot, the user does the undo action.
I noticed that after an undo action, when i try to place the marker in another position, the marker count is not reset, instead it increments
check video...

[satty-marker-undo-bug.webm](https://github.com/gabm/satty/assets/71889751/32b3db23-9b9f-4e40-b508-eeee3668c5c5)

### Solution

I figured i could declare a global mutable variable that kept count of the current marker number and decrements by 1 after every undo action.

I looked for a way to manage global variables and found that using the `lazy_static` crate provided a safer and cleaner way to manage global static variables.

I think i might have to switch to [once_cell](https://docs.rs/once_cell/latest/once_cell/)

gabm:
thank you so much for giving this a try.. but I'm afraid that this won't work on all circumstances..

a simple counter example is

- use the marker tool once - gives number 1
- use any other tool (e.g. arrow)
- undo the arrow
- create another marker - gives number 1 again!

Your solution is to create a global marker tool that is being used to track the number count - but using a local tool instance to do the "tool work"... that local instance works as before - except that it takes the number from the global instance...

Currently the architecture is, that tools create "drawables". Once the editing via the tool is finished, a drawable gets committed to the drawable stack in the sketchboard. Undo pops from the drawable stack and pushes to the redo stack, redo pops from the redo stack and pushes to the drawable stack.

By design, the sketchboard doesn't know anything about the workings of tools and drawables. That also means it doesn't know that popping a drawable may lead to side-effects such as decrementing a number. That should stay that way i think..

Without thinking it through, i believe the best approach would be to let the tool know that something has been reverted but another event.. this way the tool can then take proper action or choose to ignore it..

gabm:
I agree, that is a bug! As explained in the PR, i think we should introduce a dedicated undo/redo event that hands the drawable back to the tool and notifies it of those events

gabm:
I fixed this via a "hack".. I introduced a reference to the "next_number" counter in the tool inside the Marker and added a function to each drawable that notifies them about undo/redo events. On those events, the marker then manipulates the tools counter... I am not proud of this but it seems less invasive than other options.

[...]

PR #9, opened 2023-10-27
fix(bug): reset marker counter after undo action

noornee:
This solves issue #8 
I would appreciate your feedback and any recommendations for improving this solution. 

## video showing bug fix

[satty-marker-undo-bug_fix.webm](https://github.com/gabm/satty/assets/71889751/f62b0a8c-756d-42ef-b0a2-8465bfcced32)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,3 +23,5 @@ anyhow = "1.0"
 clap = { version = "4.4.4", features = ["derive"] }
 swayipc = { git = "https://github.com/JayceFayne/swayipc-rs.git", branch = "master" }
 libwayshot = "0.2.0"
+
+lazy_static = "1.4.0"
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -15,7 +15,7 @@ use relm4::{gtk, Component, ComponentParts, ComponentSender};
 
 use crate::math::Vec2D;
 use crate::style::{Color, Size, Style};
-use crate::tools::{Drawable, Tool, ToolEvent, ToolUpdateResult, Tools, ToolsManager};
+use crate::tools::{self, Drawable, Tool, ToolEvent, ToolUpdateResult, Tools, ToolsManager};
 
 #[derive(Debug, Clone, Copy)]
 pub enum SketchBoardMessage {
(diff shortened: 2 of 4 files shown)


gabm:
thank you so much for giving this a try.. but I'm afraid that this won't work on all circumstances..

a simple counter example is

- use the marker tool once - gives number 1
- use any other tool (e.g. arrow)
- undo the arrow
- create another marker - gives number 1 again!

Your solution is to create a global marker tool that is being used to track the number count - but using a local tool instance to do the "tool work"... that local instance works as before - except that it takes the number from the global instance...

Currently the architecture is, that tools create "drawables". Once the editing via the tool is finished, a drawable gets committed to the drawable stack in the sketchboard. Undo pops from the drawable stack and pushes to the redo stack, redo pops from the redo stack and pushes to the drawable stack.

By design, the sketchboard doesn't know anything about the workings of tools and drawables. That also means it doesn't know that popping a drawable may lead to side-effects such as decrementing a number. That should stay that way i think..

Without thinking it through, i believe the best approach would be to let the tool know that something has been reverted by introducing an additional event.. this way the tool can then take proper action or choose to ignore it..

noornee:
Thank you so much for your response and valuable feedback!

I totally understand the concern you've raised about the limitations of this solution.
I only went with using a global variable cos that was the only think i could think of at the time.

The fact that the sketchboard shouldn't be aware of tool or drawable specific details is a reasonable design principle andd I agree with the approach of introducing an additional event to notify the tool when something is reverted

I would have loved to do this myself but I'm still in the process of learning the language and am not yet confident enough to implement the proposed changes myself ;-;

gabm:
Thank you for your efforts. As explained in the issue, the fix is here: https://github.com/gabm/satty/commit/74880878c42fb7ef8fefbf2e5dcd50ccc556e91c

I am sorry to reject your PR, but it has the mentioned functional shortcummings as it stands. I still want to encourage you to continue to contribute, pls reach out to me on Matrix chat: @gabm:matrix.org - I can give you guidance next time!

noornee:
> Thank you for your efforts. As explained in the issue, the fix is here: [7488087](https://github.com/gabm/satty/commit/74880878c42fb7ef8fefbf2e5dcd50ccc556e91c)
> 
niceeeee! :star_struck: 

> I am sorry to reject your PR, but it has the mentioned functional shortcummings as it stands. I still want to encourage you to continue to contribute, pls reach out to me on Matrix chat: `@gabm:matrix.org` - I can give you guidance next time!

There's no problemo :) Thank you!
I would totally contribute whenever i see an issue i could help with andd Thanks for providing your matrix id, i would reach out whenever i need guidance! (i dont have a matrix account atm but it's about time i create one)

Issue #15, opened 2023-11-15
make install compatibility with sudo and root not having cargo


RobertMueller2:
- My cargo is installed to `$HOME/.local/bin` via rustup, root doesn't have it.
- if I want to install satty to `/usr/local`, I need root to do that.
- install depends on build-release, so `make install` includes `cargo build --release`, which root can't do.

Of course, I can use `make build-release; touch build-release; sudo make install; rm build-release` as a workaround. Or install cargo for root, of course. But it'd be nice if I didn't have to. ;)

I think the easiest is to have install depend on the actual file target/release/satty instead of build-release.

gabm:
thanks for your report. I agree that this is suboptimal.

I am not very experienced with makefiles, could you create a pr?

RobertMueller2:
Of course, happy to do so. Might take a few days before I find some spare time. :)

gabm:
fixed in #16

PR #16, opened 2023-11-16
Make install depend on target/release/satty

RobertMueller2:
PR for #15 

This should allow usage of sudo make install when root does not have cargo.

Also added force-build, force-build-release targets, because we can't check changed upstream crate dependencies in a Makefile very well.


gabm:
thank you, that was fast!

PR #18, opened 2023-11-20
fix `Cargo.lock`

a-kenji:
Previously the package was incapable of being built through the lock file.

The package can now be built using `Cargo.lock`.


gabm:
i might have forgotten that.. thank you

PR #20, opened 2023-11-29
allow sending capture to stdout

### LINKED ISSUE #19: Clipboard save doesn't seem to be picked by `wl-paste`
:wave:  thx for building satty, it's super useful.

I'm using cliphist (https://github.com/sentriz/cliphist) as a history manager and the way that its instructions recommend setting it up is via `wl-paste --watch cliphist store`

I've noticed that when using satty, `wl-paste --watch` doesn't catch things that are being copied into the clipboard by it. Not sure what's missing here but a possible solution could be to allow sending the output to stdout so I can pipe it to `wl-copy`. I just did a quick test myself by commenting this validation https://github.com/gabm/Satty/blob/main/src/sketch_board.rs#L157-L164 and then doing `satty -- output-filename /dev/stdout | wl-copy` and that worked.

marcosnils:
this is useful for piping the output to other programs, like wl-copy
as it prevents some race conditions with some programs using things like
wl-paste --watch.

Fixes #19

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -138,7 +138,7 @@ impl App {
             Some(display) => {
                 gtk::style_context_add_provider_for_display(&display, &css_provider, 1)
             }
-            None => println!("Cannot apply style"),
+            None => eprintln!("Cannot apply style"),
         }
     }
 }
@@ -277,7 +277,7 @@ fn main() -> Result<()> {
 
     match run_satty(args) {
         Err(e) => {
-            println!("Error: {e}");
+            eprintln!("Error: {e}");
             Err(e)
         }
         Ok(v) => Ok(v),
--- a/src/main.rs
+++ b/src/main.rs
@@ -224,9 +224,7 @@ impl Component for App {
 
         // Toolbars
         let tools_toolbar = ToolsToolbar::builder()
-            .launch(ToolsToolbarConfig {
-                show_save_button: config.args.output_filename.is_some(),
-            })
+            .launch(ToolsToolbarConfig {})
             .forward(sketch_board.sender(), |e| SketchBoardInput::ToolbarEvent(e));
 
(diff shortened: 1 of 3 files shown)


src/sketch_board.rs
```diff
@@ -141,41 +142,44 @@ impl SketchBoard {
...
         };
 
         let data = texture.save_to_png_bytes();
 
-        let msg = match fs::write(output_filename, data) {
-            Err(e) => format!("Error while saving file: {e}"),
-            Ok(_) => format!("File saved to '{}'.", output_filename),
+        match &self.config.output_filename {
+            None => {
+                msg = match io::stdout().write_all(&data) {
```
gabm, line 163: We should only write to stdout if it's not a tty, right?

src/main.rs
```diff
@@ -224,9 +224,7 @@ impl Component for App {
 
         // Toolbars
         let tools_toolbar = ToolsToolbar::builder()
-            .launch(ToolsToolbarConfig {
-                show_save_button: config.args.output_filename.is_some(),
```
gabm, line 228: we should only show the save button, if stdout is not a tty, correct?

gabm:
could you try to resolve the conflicts pls?

src/sketch_board.rs
marcosnils: I didn't understand that we still wanted to do this. I'll update the PR

marcosnils:
I've decided no to pursue this anymore since it needs some conflict resolution and with the `copy-command` introduced feature, that already solves this  and my use-case. I do still believe there's value in getting the output via `stdout`, but I personally don't need this anymore.

Issue #21, opened 2023-11-29
Preselect `crop` like flameshot


TD-Sky:
I use such command to capture screen area like flameshot in hyprland:

```nu
# nushell
grim -o (hyprctl monitors -j | from json | where focused | first | get name) - | satty -f - --fullscreen --output-filename $saved
```

But unlike flameshot, the default operation of satty is `Pointer` so I have to choose `Crop` and then crop the area. It takes more steps.

Are there any plans to support preselecting mode in command line or changing the default operation of satty?

gabm:
Indeed, the crop tool was default before the pointer tool was introduced. Then I figured that the pointer was a more sane default.

So I guess the easiest is to introduce an option to choose the default. I personally would also be fine to switch it to the crop tool but i am almost sure we will get this request for the other tools at some point...

Would you want to give it a shot?

gabm:
Apart from that I also want to introduced hotkeys for choosing the tools and options at some point - but that will only be a half-solution

TD-Sky:
I'm not familiar with GTK but can give a try.

TD-Sky:
Implement in this PR: https://github.com/gabm/Satty/pull/22

PR #22, opened 2023-11-29
Select tool in toolbar on startup

TD-Sky:
Implementation of https://github.com/gabm/Satty/issues/21

There are some points I need to explain:

- I defined a struct `Tools` (in `crate::command_line`) copied from `crate::tools::Tools` for the latent demand of generating completion scripts in future.
- I utilized [clap doc-comments](https://docs.rs/clap/latest/clap/_derive/index.html#doc-comments). It will strip the end `.`s of help information so I striped them manually.
- I optimize the code according to rust linter **clippy** within a commit. The new code is equivalent to the old, you don't need to spend too much time to review them.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-    "rust-analyzer.linkedProjects": [
-        "./Cargo.toml"
-    ]
+  "rust-analyzer.linkedProjects": ["./Cargo.toml"],
+  "rust-analyzer.checkOnSave.command": "clippy"
 }
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -1,28 +1,63 @@
-use clap::Parser;
+use clap::{Parser, ValueEnum};
 
 #[derive(Parser, Debug)]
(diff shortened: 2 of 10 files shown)


gabm (commented):
thank you for this pr

- the linter did interesting work, that will be helpful! thank you
- for the default tool: can't we reuse the existing tools enum for this? i think it might be better to have a single point of reference..

TD-Sky:
To generate completion scripts, a self-described command line parser which doesn't depend on other modules in crate is necessary. It would be reused in `build.rs` as a module of build script. You can reference [my code](https://github.com/TD-Sky/conceal/blob/e700703ed1b3dc2d845a60359516f12d2781991e/build.rs#L1C1-L1C1) .

Clap also allows to generate completion scripts at runtime of CLI, but compile time generation which uses `build.rs` is more convenient for CI/CD. I chose the latter.

I could make a PR for CI/CD and completion scripts this weekend if you need.

gabm:
got it.. thank you for explaining. If you feel like, i would be very happy for that contribution

PR #24, opened 2023-12-02
Generate completions and CI/CD

TD-Sky:
Testing....


TD-Sky:
Test was over.

gabm:
wow thank you, what a great contribution :tada: 

does this also install and package the icon the .desktop file?

TD-Sky:
> wow thank you, what a great contribution 🎉
> 
> does this also install and package the icon the .desktop file?

It doesn't actually do installation instead it packs the things what installation needs.

gabm:
@TD-Sky there seems to be an error with the release pipeline?! https://github.com/gabm/Satty/actions/runs/7113283418/job/19364916436

TD-Sky:
Have you retried running action?

gabm:
I can do that but it says 403 .. so I guess a token is missing or something?

gabm:
I guess something from here? https://github.com/softprops/action-gh-release#permissions

gabm:
no.. it seems that also can/needs to be requested as part of the workflow? 

```
permissions:
  contents: write
```

gabm:
https://github.com/gabm/Satty/pull/32

TD-Sky:
I don't think so. I never write that when I missed write permission problem and just turned on the setting.

gabm:
works after #32

PR #40, opened 2024-01-23
Add install instructions for alpine

gabm:
Adding installation instructions for Alpine Linux.

@sodface can you review please?


sodface:
Looks good to me!

PR #48, opened 2024-02-16
feat: save after copy

FUFSoB:
Implementing an option similar to flameshot's setting "Save image after copy"

Motivation: useful with `early-exit = true` so you can view your screenshot later without relying on clipboard manager history, or saving manually, or configuring `copy-command` solely for this


gabm:
nice, thanks! that is very useful!

gabm:
this has been released as https://github.com/gabm/Satty/releases/tag/v0.10.0

PR #52, opened 2024-02-22
Add support for non-latin alphabet

Clareses:
Use IMContext to obtain the output of the input method to support non-Latin characters.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -162,9 +162,12 @@ impl Component for App {
             // and send the messages there
             add_controller = gtk::EventControllerKey {
                 connect_key_pressed[sketch_board_sender] => move |controller, key, code, modifier | {
-                    let im_context = controller.im_context().unwrap();
-                    im_context.focus_in();
-                    if !im_context.filter_keypress(controller.current_event().unwrap()) {
+                    if let Some(im_context) = controller.im_context() {
+                        im_context.focus_in();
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -161,8 +161,12 @@ impl Component for App {
             // this should be inside Sketchboard, but doesn't seem so work there. We hook it here
             // and send the messages there
             add_controller = gtk::EventControllerKey {
-                connect_key_pressed[sketch_board_sender] => move | _, key, code, modifier | {
-                    sketch_board_sender.emit(SketchBoardInput::new_key_event(KeyEventMsg::new(key, code, modifier)));
+                connect_key_pressed[sketch_board_sender] => move |controller, key, code, modifier | {
+                    let im_context = controller.im_context().unwrap();
```
gabm, line 165: can we avoid unwrap .. it shouldn't panic imho (?)

gabm:
cool, you cracked it! great!

src/main.rs
Clareses: Sorry, I'm a Rust beginner. Is this the correct way for me to handle it?

gabm: yes, `if let Some(..) = ` is good.

Clareses: I have already modified and committed

gabm:
I will merge this into the multi-input branch.. even though there are open questions, I do think we can release that as it improves the status-quo

PR #68, opened 2024-03-25
configurable font for text annotations

ri-char:
It allows user to customize font for text annotations. Otherwise, Chinese Characters can not be rendered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
 save-after-copy = false
 # Hide toolbars by default
 default-hide-toolbars = false
+# Path of font to use for text annotations
+font = "/usr/share/fonts/TTF/Roboto-Regular.ttf"
 
 # custom colours for the colour palette
 [color-palette]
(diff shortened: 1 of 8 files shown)


gabm:
thank you so much for this contribution, I appreciate your effort...

But it makes me wonder whether it would make sense to leverage fontconfig instead? that would be more in line with other software: let the user specify a font family via the config and look it up via fontconfig.. and have a (packaged) fallback included...

README.md
```diff
@@ -78,6 +78,9 @@ output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
 save-after-copy = false
 # Hide toolbars by default
 default-hide-toolbars = false
+# Font to use for text annotations
+font-family = "Roboto"
```
gabm, line 82: I would prefer a "toml object" here

```toml
[font]
family = "Roboto"
style = "Bold"
```

What do you think?

src/femtovg_area/imp.rs
```diff
@@ -162,14 +164,34 @@ impl FemtoVGArea {
...
+                    .borrow_mut()
+                    .as_mut()
+                    .unwrap() // this unwrap is safe because it gets placed above
+                    .add_font(f.path)
+                    .ok()
+            });
+        if let Some(font) = font {
+            self.font.borrow_mut().replace(font);
+        } else {
+            self.font.borrow_mut().replace(
```
gabm, line 186: we might want to fail completely or at least show an error message?

gabm (commented):
Thank you so much for this contribution. I left some comments

gabm:
thank you!

PR #72, opened 2024-03-29
feat: add highlighter tool

DerpDays:
Currently this is an initial attempt at adding a highlighter tool, currently there are two styles that I created, however the second one (which is used in sharex) doesn't really work very well with most colors (an example of a good color is #FFFF00), however i've included it incase you prefer it.

First Style (better imo):
satty_style_1
Second Style:
satty_style_2

However, there is an issue with my current implementation, which is that to apply the filter I take a screenshot of the canvas and then create a new image with the filter applied from the pixels from the screenshot, however `canvas.screenshot()` does not result in a full resolution image, meaning that the result is blurry if the highlight is applied when the size of satty is a different resolution. And this overall is not a great solution. Also if the satty is open in a different aspect ratio it also is misaligned due to the coordinates being relative to the actual screenshot with no way to get the offset (that I found).

I tried looking for a while but didn't notice any way to get a full resolution image to be able to apply the highlight, so I was wondering if you have any ideas on other approaches, and also your opinion on which highlighter you prefer.

### DIFF
--- a/icons.toml
+++ b/icons.toml
@@ -5,6 +5,7 @@ icons = [
   "cursor-regular",
   "number-circle-1-regular",
   "drop-regular",
+  "highlight-regular",
   "arrow-redo-filled",
   "arrow-undo-filled",
   "save-regular",
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -64,6 +64,7 @@ pub enum Tools {
     Text,
     Marker,
     Blur,
+    Highlight,
     Brush,
 }
 
@@ -79,6 +80,7 @@ impl std::fmt::Display for Tools {
             Text => "text",
             Marker => "marker",
             Blur => "blur",
+            Highlight => "highlight",
             Brush => "brush",
         };
         f.write_str(s)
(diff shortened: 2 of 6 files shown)


gabm:
Thank you for your contribution!

Am I seeing this currently that you do the alpha channel mixing manually?

Isn't it the same as drawing a rectangle that has a colour with an alpha channel on top - just like what the rectangle tool does?

gabm:
I just tested it and it looks fine.. I'll ask the original issue author for review as well.. 

Notes for the future:

- this might also be handled by the rectangle tool with a transparent fill (?)
- to imitate a more "marker-like" feel, we might also draw a fixed size rectangle by only evaluating the x-component of the mouse movement
  - we might even draw a "free-hand" area with fixed height. For better looks, we can lock that free hand to "straight" when pressing SHIFT
- we might want to enhance the contrast after drawing since alpha mixing reduces it. That could be done via a histogram-based approach (?!)

gabm:
I'll merge that.. if @sdooweloc wants to give feedback, we can still incorporate that later

PR #75, opened 2024-04-23
fix gnome dash running app icon

GrzegorzKozub:
before:

Screenshot from 2024-04-23 08-50-45

after:

Screenshot from 2024-04-23 08-51-27


gabm:
well spotted, thanks!

PR #77, opened 2024-04-25
feat: draft freehand highlighter

DerpDays:
This is will add a freehand highlighter as the default (with shift to align it), and then have a selection highlight (like there currently is) accessible by pressing ctrl whilst on the highlighter, since this was suggested at some point (though since currently the highlighter tool is the selection it might make more sense to make the freehand one accessible through ctrl instead), currently this is just draft and needs quite a bit of cleanup.

EDIT: kept the block/selection highlighter as the default, with the freehand/line highlighter being the alternative highlighter, this is configurable via the CLI or config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,6 @@ apk add satty
 
 You can download a prebuilt binary for x86-64 on the [Satty Releases](https://github.com/gabm/satty/releases) page.
 
-
 ## Usage
 
 Start by providing a filename or a screenshot via stdin and annotate using the available tools. Save to clipboard or file when finished. Tools and Interface have been kept simple.
@@ -78,6 +77,9 @@ output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
 save-after-copy = false
 # Hide toolbars by default
 default-hide-toolbars = false
+# Whether to set block or line/pen as the default highlighter, other mode is accessible using CTRL.
+default-block-highlight = true
+
 # Font to use for text annotations
 [font]
 family = "Roboto"
@@ -122,6 +124,8 @@ Options:
           After copying the screenshot, save it to a file as well
   -d, --default-hide-toolbars
           Hide toolbars by default
+      --default-line-highlight
+          Change the default highlighter to the line/pen highlighter
       --font-family <FONT_FAMILY>
           Font family to use for text annotations
       --font-style <FONT_STYLE>
@@ -139,7 +143,6 @@ You can bind a key to the following command:
 ```
 grim -g "$(slurp -o -r -c '#ff0000ff')" - | satty --filename - --fullscreen --output-filename ~/Pictures/Screenshots/satty-$(date '+%Y%m%d-%H:%M:%S').png
 ```
-
 
 ## Build from source
 
--- a/README.md
+++ b/README.md
@@ -77,8 +77,8 @@ output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
 save-after-copy = false
 # Hide toolbars by default
 default-hide-toolbars = false
-# Whether to set block or line/pen as the default highlighter, other mode is accessible using CTRL.
-default-block-highlight = true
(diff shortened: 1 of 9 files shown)


00-kat:
> though since currently the highlighter tool is the selection it might make more sense to make the freehand one accessible through ctrl instead

I think (not a dev, just a random user) that it should probably be a config option you can set in the config file.

DerpDays:
yeah that's probably best, i'll get to working on this in a few days so it can finally not be a draft

gabm:
I agree.. I didn't have a look at the PR, @DerpDays let me know if I can help

DerpDays:
Sorry for the large delay & also the large commit (this should have been broken down into quite a few more commits).

This should mostly ready, here's an overview of the new things:
- Adds a new highlighter, which is a freehand line/pen highlight, by default it is the secondary highlighter meaning it is accessible with CTRL, by pressing SHIFT while highlighting it aligns it from the previous point that was not aligned. To make it easier to make perfect turns, you can release SHIFT to add an extra point (and do it again to remove the point if you want) when you haven't moved the cursor.
- Removed the border around the block highlighter since its not necessary and inconsistent with other tools apart from the blur tool.
- Added a SHIFT option for the block highlight to make a square highlight.
- Configuration/command line options to change the primary highlighter to the line highlighter.
- Add a new event for tools: key_release which as implied is when a key is released (in contrast to the previous event of when pressed)

Let me know if there's anything that needs changing or can be done better.

00-kat:
If you start holding shift mid-way should it draw a straight line from the beginning or from the spot you started holding shift from (it does latter rn)?

DerpDays:
The spot you started holding shift from, thought it'd give more options for how people want to draw it, since you can just hold shift when you start the highlight to achieve the other effect

gabm:
I just tested it, it looks great!

src/tools/highlight.rs
```diff
@@ -48,75 +87,169 @@ impl Drawable for Highlight {
...
+        let default_highlight_block = APP_CONFIG.read().default_block_highlight();
         match event.type_ {
             MouseEventType::BeginDrag => {
-                self.highlight = Some(Highlight {
-                    top_left: event.pos,
-                    size: None,
-                    style: self.style,
-                    editing: true,
-                });
+                match (ctrl_pressed, default_highlight_block) {
```
gabm, line 130: Seeing this "complex" logic, maybe it would be good to add some lines of comment somewhere in this file about the interaction model

- two modifier keys (ctrl, shift) trigger two different actions (block, free-hand highlight)
- free-hand highlight can perform two different things: freehand and "straight line"
- ...

gabm (changes requested):
Please adress the comments above, then we are good to go! Great contribution :tada:

README.md
```diff
@@ -122,6 +124,8 @@ Options:
           After copying the screenshot, save it to a file as well
   -d, --default-hide-toolbars
           Hide toolbars by default
+      --default-line-highlight
+          Change the default highlighter to the line/pen highlighter
```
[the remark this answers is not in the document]
DerpDays: I've changed it now to be a string option, I'm unsure if this is the way we want to do it, or if we stick to having it be a bool.

I had originally changed the CLI wording since clap didn't want to parse `--default-block-highlight=n`, however taking a look, I'm assuming that would've worked if I made it optional and then set it to be valueparsed with the bool value parser.

Let me know if you prefer having it as a string/enum or keeping it as an option, which with the new consistent naming would be along the lines of: block-highlighter-primary

src/tools/highlight.rs
DerpDays: I was thinking that these explanations probably deserves to be somewhere on the README. What would you think of having them in the config (since it's shown on the README), I'd have try to make it concise though

gabm: that sounds good!

README.md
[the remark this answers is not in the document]
gabm: I like your solution here!

gabm (approved):

gabm:
Thank you for this great contribution :partying_face:

PR #79, opened 2024-05-01
Change /use to /usr in README

cd-CreepArghhh:
*Incredibly* minor fix. It was kinda annoying me :laughing:


gabm:
🤦‍♂️

good catch!

PR #96, opened 2024-07-13
add short cli option for --output-filename

jacekpoz:


gabm:
makes a lot of sense!

PR #98, opened 2024-07-13
add ellipse tool

jacekpoz:
the current behaviour is the middle of the circle stays where you started dragging

I could change it so that it functions similar to the rectangle but it's 1 in the morning and I can't really think

if you do want that I'll do it in the morning


gabm:
looks good.. 

do you think you can add the shift modifier to draw circles (`radii.x = radii.y`)?

jacekpoz:
I'll also add the same thing to the rectangle tool in another PR

PR #99, opened 2024-07-13
add an option to fill shapes

jacekpoz:
these changes might be rough in some places as I don't know most of the libraries you're using very well

I mostly copied other parts of the code and partly read the docs if needed

the ellipse tool PR will also need a small change to support this, I'll add it when you merge that one


gabm:
do you want to add support for the ellipse tool after rebasing?

jacekpoz:
there, that should be all?

PR #102, opened 2024-07-14
flake: remove numtide/flake-utils

jacekpoz:
this flake's usage of flake-utils is trivial enough where we can replace it with a few lines of boilerplate

this gives us the benefit of having one less input and removing a lot of hidden complexity and potentially another copy of flake-utils from the devs' computers

this is also a great post on the matter: https://ayats.org/blog/no-flake-utils

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,22 +4,24 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     rust-overlay.url = "github:oxalica/rust-overlay";
-    flake-utils.url = "github:numtide/flake-utils";
   };
 
-  outputs = { self, nixpkgs, rust-overlay, flake-utils, ... }:
-    flake-utils.lib.eachDefaultSystem (system:
-      let
-        overlays = [ (import rust-overlay) ];
-        pkgs = import nixpkgs {
-          inherit system overlays;
-        };
-
-      in
-      with pkgs;
-      {
-        devShells.default = mkShell {
-          buildInputs = [
+  outputs = {
+    nixpkgs,
+    rust-overlay,
+    ...
+  }: let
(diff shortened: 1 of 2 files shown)


NotAShelf:
+1 for removing flake-utils! It's a relic from a less civilized era

flake.nix
```diff
@@ -28,16 +30,16 @@
...
-              {
-                extensions = [ "rust-src" ];
-              })
+            (rust-bin.stable.latest.default.override {
+              extensions = ["rust-src"];
+            })
           ];
 
-          shellHook = ''
+          shellHook = with pkgs; ''
```
diniamo, line 38: Don't use `with pkgs` here

flake.nix
```diff
@@ -4,22 +4,24 @@
...
+    rust-overlay,
+    ...
+  }: let
+    systems = ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"];
+    forEachSystem = nixpkgs.lib.genAttrs systems;
+  in {
+    devShells = forEachSystem (
+      system: let
+        overlays = [(import rust-overlay)];
+        pkgs = import nixpkgs {inherit system overlays;};
```
diniamo, line 20: Don't import `nixpkgs`, use `nixpkgs.legacyPackages.${system}` instead. `rust-overlay` also exposes its packages via `rust-overlay.packages.${system}`.

See https://zimbatm.com/notes/1000-instances-of-nixpkgs, if you want a detailed reasoning.

diniamo (changes requested):
Other than that, +1, flake-utils bad >:(

gabm:
thanks .. extra 👍 for the link!

diniamo:
...

gabm:
@diniamo i didn't see your review before merging - sorry!

PR #107, opened 2024-07-28
feat: option to disable notifications

### LINKED ISSUE #80: [Feature Request] Option to Disable Notification
When you copy (or save I think) a screenshot, a notification chimes in:
image
Could there be an option (maybe `--disable-notification`) to disable the notification from showing?

Also, may I ask why the notification exists in the first place? Since I feel like it would be incredibly easy to use
```
satty -f - && notify-send -i /usr/share/icons/hicolor/scalable/apps/satty.svg 'Satty' 'Copied to clipboard.'
```
when you call it (from maybe a keybind or something) and to maybe even create a `satty` function that runs `notify-send` after if you use it from a shell.

danielwerg:
Closes #80 

Please review @gabm

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ save-after-copy = false
 default-hide-toolbars = false
 # The primary highlighter to use, the other is accessible by holding CTRL at the start of a highlight [possible values: block, freehand]
 primary-highlighter = "block"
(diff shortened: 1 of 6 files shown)


gabm:
great, thank you! @danielwerg

PR #108, opened 2024-07-28
Update README.md

sodface:
Update Alpine installation to reflect that the package moved from  testing to community


gabm:
great news!

PR #110, opened 2024-08-03
Add Panning and Zooming

### LINKED ISSUE #93: [feature] Allow for pan and zoom with the mouse/trackpad to move the (infinite) canvas around
I was going to ask to have zooming added because it's hard to annotate accurately when you have a huge screenshot (like 2 screens in my case) and I found this was [already discussed](https://github.com/gabm/Satty/issues/49#issuecomment-1959012884). The issue where it is mentioned (#49) is for something else and I would prefer this feature (pan/zoom) doesn't get lost.

> The cool thing would be that it brings a ton of improvements
>
> - if something is hidden behind the toolbars, you can just move until its visible
> - you can draw in and outside the image (infinite canvas)
> - you can zoom in to work on a very small part of the image
> - you annotate very small or very big stuff

mattrobineau:
## Description

This PR introduces changes to allow a user to pan and zoom the canvas in Satty and closes #93.

## Changes made:

- Adds zooming functionality
  - Adds two toolbar buttons for zoom in and zoom out
  - Adds scrollwheel up/down zooming
- Adds panning functionality
  - Holding and dragging middle mouse button will pan the image 

## Notes
- The scroll event `gtk::EventControllerScroll` needed to be added to the main window otherwise the event would only fire when scrolling on the toolbar.
- `handle_zoom` zooms by an arbitrary value (+-0.1) chosen solely because I thought it was a good value. It might be worth considering allowing a user to modify this value in config.
- Zooming does not currently support zooming based on the cursors location, it will always zoom from (0,0).


gabm:
I just tried your PR and the functionality works as advertised, great :+1: 

For the current feature-set I'd like to suggest the following changes

# zoom

- I think the center of zooming should be:
  - a) the center of the screen, if invoked by the toolbar buttons
  - b) the position of the mouse pointer, if invoked by the scroll wheel
- to change the center of the zooming operation, you need to translate the chosen center to the left-top corner, then zoom and then translate back

# pan

- the pan speed should be equal to the mouse speed
- on my screen it seems that the canvas moves "slower" than the mouse moves
- that might have something to do with the scale factor (?)

---

# Usability and UI

- we need to discuss whether we want toolbar buttons, an extra toolbar or none 
- we may want to respond to the arrow buttons for panning
- I think we want a shortcut to reset all zoom and translation
- we need to think how the background looks like - black or white (or any other color)
- we do want to export/save the "smallest possible canvas", that still covers the image and all drawings. That allows users to draw outside the image
- I didn't look at the code yet :see_no_evil:

FelipeSharkao:
Is there any blocker for this to merged? I've looking forward for this feature.
I could take a shot at fixing any problems in this changes if @mattrobineau is not able rn, since I'm interested in get this working

mattrobineau:
I don't currently have the bandwidth to continue working on this. If you would like to take over, please feel free.

The current state of this PR adds working panning however the zooming is not currently working correctly. The zoom button works fine, however the zoom under cursor is not completed.

RobertMueller2:
This inspired/was superseded by #336. So I'm closing this one.

PR #119, opened 2024-09-05
Execute `copy-command` via system shell

### LINKED ISSUE #105: "No such file..." error on copy to clipboard
I receive the following output when I attempt to copy to clipboard: 
> ```
> ~> grim - | satty --filename - --fullscreen
>
> (process:148659): Gtk-WARNING **: 22:57:10.148: Unknown key gtk-modules in /home/vegard/.config/gtk-4.0/settings.ini
> 
> (satty:148659): Adwaita-WARNING **: 22:57:10.174: Using GtkSettings:gtk-application-prefer-dark-theme with libadwaita is unsupported. Please use AdwStyleManager:color-scheme instead.
> Error saving No such file or directory (os error 2)
> ```

Saving files works just fine. I've also tried running it with an image path instead of piping to it.

Other scenarios that successfully copies the image to clipboard:
```shell
# plain grim + wl-copy
grim - | wl-copy -t image/png

# swappy works as well
grim - | swappy -f -
```

I use satty 0.13.0, with wl-clipboard. 

In addition, I have copyq set up to manage my clipboard history (I've also tried without it running - just in case).

I've set `copy-command = "wl-copy -t image/png"` in the config.

LRitzdorf:
...specifically, `sh`. This allows for commands with arguments, such as `wl-copy -t image/png`, or even more advanced shell features.

This technically also allows for arbitrary command execution, but the user already has execution privileges anyway if they're providing a copy command.

Fixes #105

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -232,7 +232,9 @@ impl SketchBoard {
         texture: &impl IsA<Texture>,
         command: &str,
     ) -> anyhow::Result<()> {
-        let mut child = Command::new(command)
+        let mut child = Command::new("sh")
+            .arg("-c")
+            .arg(command)
             .stdin(Stdio::piped())
             .stdout(Stdio::null())
             .spawn()?;


gabm:
oh thats great @LRitzdorf, thank you for hunting this down.

gabm:
.. unfortunately it doesn't compile .. could you check?

LRitzdorf:
...oh, duh, `args()` is not the same as `arg()`. Fix incoming.

LRitzdorf:
FYI, this does indeed compile now, and I've gotten around to testing the full executable on my system, where it works as expected!

gabm:
great thanks!

Bykow:
Any clue when this will be released ?

PR #124, opened 2024-09-25
Update a command for installing Satty in Arch Linux

### LINKED ISSUE #122: satty-bin no longer in AUR
The satty-bin package is no longer available from the arch user repository, but the readme still shows it as the package to install. Was this intentional?

TheBits:
Hi @gabm. Thank you for Satty! I've fix #122


gabm:
thank you!

PR #139, opened 2024-12-19
feat: make color configuration more flexible

### LINKED ISSUE #135: [feature] option to configure multiple custom colors
Currently, only one custom color may be set via the configuration file (`color-palette.custom`). It would be useful to have this be a list instead, where the first element of that list would be the color shown in satty's own UI, and the rest (maybe including the first) could be the presets at the bottom of the GTK color picker (see `gtk_color_chooser_add_palette`).

Additionally, it might be worth considering to refactor the whole color configuration so the user can set an arbitrary amount of colors for the color palette.

Edit: Please let me know, if I should open a PR implementing this!

LordMZTE:
This refactors the color configuration options such that the user can specify an arbitrary amount of colors for the toolbar, and also specify colors as presets for the color picker. I took the liberty to implement this without prior feedback, sorry :D
image

Closes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,14 +86,28 @@ disable-notifications = false
 family = "Roboto"
 style = "Bold"
 
-# custom colours for the colour palette
+# Custom colours for the colour palette
 [color-palette]
-first= "#00ffff"
-second= "#a52a2a"
-third= "#dc143c"
-fourth= "#ff1493"
-fifth= "#ffd700"
-custom= "#008000"
+# These will be shown in the toolbar for quick selection
+palette = [
+    "#00ffff",
+    "#a52a2a",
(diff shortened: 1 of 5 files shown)


gabm:
lets wait for the checks, then I'll merge

gabm:
@LordMZTE there is a small complaint from the linter

LordMZTE:
Thanks! I fixed the lint and also updated the nix flake (`nix flake update`), as the clippy version of the nix dev shell didn't catch this.

w-jablonski:
In the "Choose Color" dialog, the "Cancel" button does nothing when it should close the dialog... maybe I should open a bug...  Satty 0.19.0 Arch Sway

PR #140, opened 2024-12-19
option to specify the "corner roundness"

schaumtier:
new option "corner-roundness". "0" means no corners. "12" is the default value (which draws corners of rectangle, blur and highlight round.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ fullscreen = true
 # Exit directly after copy/save action
 early-exit = true
 # Draw corners of rectangles round if the value is greater than 0 (0 disabled rounded corners)
-corner_roundness = 12,
+corner-roundness = 12
 # Select the tool on startup [possible values: pointer, crop, line, arrow, rectangle, text, marker, blur, brush]
 initial-tool = "brush"
 # Configure the command to be called on copy, for example `wl-copy`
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -65,6 +65,8 @@ All configuration is done either at the config file in `XDG_CONFIG_DIR/.config/s
 fullscreen = true
 # Exit directly after copy/save action
 early-exit = true
+# Draw corners of rectangles round if the value is greater than 0 (0 disabled rounded corners)
+corner_roundness = 12,
```
LordMZTE: There shouldn't be a comma here. We should also consider writing `12.0` to clarify that this is a float, not an int.

src/style.rs
```diff
@@ -215,11 +224,12 @@ impl Size {
         }
     }
 
-    pub fn to_corner_radius(self) -> f32 {
+    pub fn to_corner_roundness(self) -> f32 {
```
LordMZTE, line 227: Why is this function needed? It always returns the same value, regardless of `self`. At a glance, it also seems unused.

src/style.rs
```diff
@@ -18,6 +18,9 @@ pub struct Style {
     pub fill: bool,
 }
 
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Properties {}
```
LordMZTE, line 22: Why do we need this struct?

src/style.rs
```diff
@@ -176,6 +179,12 @@ impl FromVariant for Size {
     }
 }
 
+impl Properties {
+    pub fn corner_roundness() -> f32 {
```
LordMZTE, line 183: This seems a little redundant. Perhaps we could directly reference `APP_CONFIG.read().corner_roundness()` where it's used.

LordMZTE (changes requested):
Thanks! I left some comments on things I believe could be improved.

README.md
schaumtier: removed it already

src/style.rs
schaumtier: i don't really know - just needed a place to attach the getter function... maybe its better kept somewhere else. rust is really not my programming language

src/style.rs
schaumtier: initially i thought of the need to differentiate between "Small", "Medium" and "Large" - thats why i've made a function. but of course would it be enough to have the numeric value.
if it's okay to reference APP_CONFIG directly in e.g. rectangle.rs, i can do this there.

LordMZTE (approved):

gabm:
great, thank you both!

PR #143, opened 2024-12-28
Draw a fuller arrow

bobvanderlinden:
The current arrow is composed of 3 thin lines. It would be nice to have a more fatter arrow, so that the individual lines of the arrow aren't as prominent.

The arrow looked like:

image

(from small to large, unfilled first, then filled)

Now they look like:

image

### DIFF
--- a/src/math.rs
+++ b/src/math.rs
@@ -10,6 +10,38 @@ pub struct Vec2D {
     pub y: f32,
 }
 
+#[derive(Default, Debug, Copy, Clone, PartialEq)]
+pub struct Angle {
+    pub radians: f32,
+}
+impl Angle {
+    pub fn from_radians(radians: f32) -> Self {
+        Self { radians }
+    }
+
+    pub fn from_degrees(degrees: f32) -> Self {
+        Self {
+            radians: degrees * PI / 180.0,
+        }
+    }
+
+    pub fn cos(&self) -> f32 {
+        self.radians.cos()
+    }
+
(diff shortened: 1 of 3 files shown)


gabm:
I like the looks imho, code looks good to me! Would be a nice addition.

Currently I see these things

- fixing the bug you discovered
- can you post a before -after image?
- doc: 
  - looking this PR
  - adding an ASCII drawing as glossary (nice to have)

gabm:
I like the looks imho, code looks good to me! Would be a nice addition.

Currently I see these things

- fixing the bug you discovered
- can you post a before -after image?
- doc: 
  - looking this PR
  - adding an ASCII drawing as glossary (nice to have)

bobvanderlinden:
I've updated the PR:

* Updated the description with before and after screenshots.
* Added better description in the comments with an attempt at ASCII-art representation :sweat_smile:
* Made sure the styles are being used by the arrow drawing. Filled style is the new fat arrow. The unfilled style uses the separate lines approach. The sizes for the two styles are being reused.

This does change the sizes of the unfilled style a bit compared to what was used previously, but it is consistent with filled style this way. Let me know what you think!

gabm:
I like the result, looks good to me! 

If u feel that it's ready, please mark as ready and I'll merge.

bobvanderlinden:
Awesome 😄 Sorry, forgot this was still in draft 😅 It's ready now!

PR #144, opened 2024-12-29
Action (save) on Enter

pol-rivero:
Other screenshot GUIs like Flameshot allow pressing the Enter key to quickly accept the current selection and save (and possibly close if configured to do so). This PR adds this behavior.

The user can choose which action to take with the new config "action-on-enter" (possible values: `save-to-clipboard`, `save-to-file`).
The default is `save-to-clipboard` (I can change it if you prefer). I also considered making it an Option and defaulting to None (doing nothing when pressing Enter unless configured), but I think it's more intuitive to have the Enter key do something by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ copy-command = "wl-copy"
 annotation-size-factor = 2
 # Filename to use for saving action. Omit to disable saving to file. Might contain format specifiers: https://docs.rs/chrono/latest/chrono/format/strftime/index.html
 output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
+# Action to perform when the Enter key is pressed [possible values: save-to-clipboard, save-to-file]
+action-on-enter = "save-to-clipboard"
 # After copying the screenshot, save it to a file as well
 save-after-copy = false
 # Hide toolbars by default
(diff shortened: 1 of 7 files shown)


pol-rivero:
It seems that I forgot to run `cargo fmt`. It should now be fixed.

pol-rivero:
I see.
When I was testing the changes, I tried each tool and the only one that seemed to react in any way to the Enter key was the Text tool (currently it commits the changes, maybe in the future it would insert a newline?), and in that case it worked as expected (the first Enter would commit the changes and the second Enter would take the screenshot).

If the Unmodified result is not reliable for determining if the tool handled the event, I'm not sure on how to best implement this. How would you do it?

> I am also not sure that this makes any difference at this point?!

I don't understand what do you mean.

gabm:
What i want to say with that is: I don't think, that at the moment this is a bug. It's okay to implement it the way you did and probably the only way, until we have proper signalling for state change, visual change and input handling.

please leave a comment at that line for now that indicates, that Unmodified is probably not the right signal to listen to, but that this is the only way at the moment.

pol-rivero:
Sure! I've added the comment.

PR #145, opened 2024-12-29
Crop improvements

pol-rivero:
Improvements:

- NEW: Dragging outside the area will create a new crop (previously did nothing)
    - Useful for starting again after a misclick

https://github.com/user-attachments/assets/bfdc24ff-4470-480f-8661-2bf29efa2183

- Dragging outside the area (but close to the region) will search the closest handle and act as if that handle had been clicked
    - Prevents accidentally losing the current selection (accidentally starting a new crop)

https://github.com/user-attachments/assets/581c47aa-a6a0-49f7-abac-5e90ddab5a29

- Dragging inside the area moves it (unchanged behavior)

- NEW: Dragging inside the area while very close to a handle will act as if that handle had been clicked
    - Prevents losing the area placement when misclicking a handle

https://github.com/user-attachments/assets/c42d0793-03f4-4629-a245-07e15ef133f2

- By clicking the area and moving it, it is possible to have a part of the area outside the bounds of the image. This previously resulted in a black background in the out-of-bounds area, while now it is excluded.

https://github.com/user-attachments/assets/14d073a1-85ff-460d-90e3-83b0b2cb5247

| Old | New |
|---|---|
| n6 | n5 |

- Due to the use of floats, using the crop tool could result in the image looking blurry. The crop is now rounded to the closest integer to avoid that issue.

| Old | New |
|---|---|
| n8 | n7 |
| image | image |

- The PR also has some minor code cleanup in `crop.rs`, mostly to move some methods that made more sense in Crop than CropTool and to make Crop.size not an Option (didn't seem necessary and also caused some minor flickering).

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -283,7 +283,7 @@ impl FemtoVgAreaMut {
             .crop_tool
             .borrow()
             .get_crop()
-            .and_then(|c| c.get_rectangle())
+            .map(|c| c.get_rectangle())
             .unwrap_or((
                 Vec2D::zero(),
                 Vec2D::new(
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -19,7 +19,7 @@ use relm4::{gtk, Sender};
 use resource::resource;
 
 use crate::{
-    math::Vec2D,
+    math::{rect_ensure_in_bounds, Vec2D},
     sketch_board::{Action, SketchBoardInput},
     tools::{CropTool, Drawable, Tool},
     APP_CONFIG,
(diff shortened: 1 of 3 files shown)


gabm:
Great improvements and fixes! 👍 

I don't remember, why size was an option back then, but now it's not anymore.. any idea why this is okay that way?

pol-rivero:
I also don't understand why it was an option. The only behavior change that I could find is that, in the first frame after clicking (before the first drag update), the crop area would be None and therefore be treated as the entire screen, while now that area has a size of 0.
However, this doesn't matter in practice, since it's practically impossible to save the screenshot in such a small period of time, so avoiding the flickering and having simpler code is worth it in my opinion.

gabm:
aaah yes, that was a real problem. it was provoked by starting a crop but not moving the mouse. Easiest achieved with a trackball, clicking on one place without moving. 

We should account for that. We can do that by checking against an epsilon, like the handle size - or with an option or similar.

pol-rivero:
Aah, ok, I've added it back.

PR #146, opened 2025-01-02
Add toolbar keyboard shortcuts

gabrielgcma:
Hi folks, this is my very first community PR ever, and I'm also doing this to learn Rust, so bear that in mind.

## Description
This PR introduces changes to allow tool selection via keyboard shorcuts and closes https://github.com/gabm/Satty/issues/50.

| Tool      | Shortcut |
|-----------|----------|
| Pointer   | Alt+P    |
| Crop      | Alt+C    |
| Brush     | Alt+B    |
| Line      | Alt+L    |
| Arrow     | Alt+A    |
| Rectangle | Alt+R    |
| Ellipse   | Alt+E    |
| Text      | Alt+T    |
| Marker    | Alt+M    |
| Blur      | Alt+U    |
| Highlight | Alt+H    |

This is the event flow I came up with. The mouse click also updates the toolbar UI `active_button` so that you don't have two buttons in the UI selected at the same time (one from keyboard and another from a click).

satty-20250102-01:36:50

## Changes made:
Adds keyboard shortcuts functionality to the editing tools

## Notes
- I took the opportunity to refactor the input keys handling on `sketch_board.rs`.
- The `tool_shortcuts` hashmap was defined inside the `update()` method of the `SketchBoard` component for skill issue reasons. I'd like to know how to improve that.
- I'm sure there are many improvements to be made and I don't know if my approach is the best, I'm a total Rust noob. Upon testing everything seems fine, but I'd be surprised if no bugs were introduced. I'd like to know your thoughts on this.

PR #147, opened 2025-01-07
feat: make layout-independent shortcuts

### LINKED ISSUE #121: Make shortcuts independent to layouts
Hi y'all!

I met a problem with shortcuts like Ctrl-C, Ctrl-T and another on layouts other than English because in code gtk4 uses keyval on specific layout. For example, the keyval on any English layouts for letter 'c' returns always `99`, but in Belarusian layout - `1747`.

I tried to figure out with key codes which are fixed, but can't find any information about it and I don't want to enter into unsafe gtk4 code. So... I gave up with it.

Any idea to fix it?

JarKz:
### Related issues

Closes #121 
Should resolve #146

#### Changes

- Added crate 'keycode' as dependency which helps us to use evdev keycodes.
- Implemented evdev keycode matching as fallback for shortcuts.

As maintainer said that the application should focus only to linux, so this changes on linux should work as remains, even on X11 (my friend has tested, but I'll accepts more tests). To match key uses `Key` from gtk4 as primary and fallbacks to use keycode which on linux is evdev keycode. That's why I've used the 'keycode' crate which have a list of evdev keycodes. This should make the shortcuts layout-independent.

#### Is it ready to merge?

I should make attention to **fallback** word. Maybe it is not correct way to use independent-layouts as default. I've asked in issue, but not got any answers and at this time I had draft code. So I implemented and can make changes if need to use config property.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,6 +88,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
 
 [[package]]
+name = "arraydeque"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f0ffd3d69bd89910509a5d31d1f1353f38ccffdd116dd0099bbd6627f7bd8ad8"
+
+[[package]]
+name = "arrayvec"
+version = "0.4.12"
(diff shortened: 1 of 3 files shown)


gabm:
I think we can go ahead with this like it is.. For current users i don't see a huge downside.. 

That being said, I can't estimate the consequences. If users start reporting bugs i am ready to revert this. Until then: let's go!

Issue #50, opened 2024-02-20
Additional keyboard shortcuts


alxndr13:
First of all: thanks for the nice software!

would be nice to support some more keyboard shortcuts. I thought about the following:

- `T` to switch to the text tool
- `A` to switch to the arrow tool
- `L` to switch to the line tool
- `R` to switch to the rectangle tool

alxndr13:
> Yep.. already had this in mind.. wasn't sure yet whether to use letters or numbers (according to the position in the top toolbar)

One argument for the letters would be that swappy also uses the letters as shortcuts :)

KAGEYAM4:
Also letters are good mnemonics for remembering, if numbers are used, will always have to look which tools is in which position.

dpi0:
> Yep.. already had this in mind.. wasn't sure yet whether to use letters or numbers (according to the position in the top toolbar)

yup! this is exactly what i was missing coming from sharex

00-kat:
> wasn't sure yet whether to use letters or numbers

Another argument for letters (a bit late): if there is ever a 11ᵗʰ item, it would be inaccessible.

tukusejssirs:
I would say: choose some defaults but make them configurable. Also add a way to use the position number as a shortcut. 😅

I prersonally would prefer configurable letters though.

fabienjuif:
I think if we are going this road we should have a matrix mapping.
If we can edit tools position (and visibility) later we will be happy to have a matrix mapping.

**Example**
- I configure the bar to be 
  - top [arrow, blur, rectangle]
  - bottom [red, blue, s|m|l]
- I want a mapping being maybe
  - [1, 2, 3] 
  - [ctrl+1, ctrl+2, ctrl+3]
- Then my configuration (matrix) would be
  - shortcut_keys_top: ["1", "2", "3"]
  - shotcut_keys_bottom: ["ctrl+1", "ctrl+2", "ctrl+3"]

And if I change my layout later because I prefer the line than the arrow, I just change the UI, and the matrix still works.

PR #148, opened 2025-01-22
feat: single-key and number-key toolbar shortcuts

gabrielgcma:
Hi folks, this is my very first community PR ever, and I'm also doing this to learn Rust, so bear that in mind.

## Description
This PR introduces changes to allow tool selection via keyboard shorcuts and closes https://github.com/gabm/Satty/issues/50.

| Tool      | Shortcut |
|-----------|----------|
| Pointer   | P   |
| Crop      | `C`  or `1` |
| Brush     | `B` or  `2`  |
| Line      | `L` or `3`  |
| Arrow     | `A` or `4`   |
| Rectangle | `R` or `5`   |
| Ellipse   | `E` or `6`   |
| Text      | `T` or `7`   |
| Marker    | `M` or `8`    |
| Blur      | `U` or `9`    |
| Highlight | `H` or `0`    |

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,8 +29,8 @@ mod style;
 mod tools;
 mod ui;
 
-use crate::sketch_board::TextEventMsg;
-use crate::sketch_board::{KeyEventMsg, SketchBoard, SketchBoardInput};
+use crate::sketch_board::{SketchBoard, SketchBoardInput};
+use crate::tools::Tools;
 
 struct App {
     image_dimensions: (i32, i32),
@@ -43,6 +43,7 @@ struct App {
 enum AppInput {
     Realized,
     ToggleToolbarsDisplay,
+    ToolSwitchShortcut(Tools),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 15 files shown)


danielwerg:
Would be nice to have `[keybinds]` section in config to customize it.

gabm:
thank you for your contribution.. does IM Context still work? e.g. Chinese input?

RobertMueller2:
I really like the idea of one key shortcuts to tools. That said, I see potential clashes with future zooming/panning functionality:

Shift 1-5 and 1-5 are used for zoom in gimp, we could reuse those for zooming. But on the other hand, Ctrl+/Ctrl-- might just be good enough.
WASD and HJKL might be familiar for moving around, we might use that to move around in a zoomed image. Not sure we need both, the latter doesn't lend itself to moving the canvas, I think, but maybe the cursor even though admittedly it has limited usefulness for sure.

I'd probably do away with the number shortcuts for now and leave that for (later) remapping via config file if users want to use those for tools.
- Arrow A-> Z (no better idea, sorry)
- hiGhlight H->G
- lIne L->I

Yes, sadly it is not as intuitive, but I think it's less painful than having to move them later.

Sorry for the remarks at this late stage when the PR was already here for a while. Thoughts? Feel free to tell me to get out of here ;)

malparty:
What about we unlock this PR with:
1. Remove numbers at the moment (until further discussions about zoom handling can progress)
2. Choose between HJKL or WASD, but IMO it's wasting keys to offer 3 different arrow-shortcuts (arrows, HJKL and WASD...)
3. Use whatever letters that fit and we can work on another PR to bring these letters to the config file :bulb: 

IMO it would bring a lot of value to start with just these few shortcuts. :bow:

RobertMueller2:
> What about we unlock this PR with:
> 
>     1. Remove numbers at the moment (until further discussions about zoom handling can progress)
> 
>     2. Choose between HJKL or WASD, but IMO it's wasting keys to offer 3 different arrow-shortcuts (arrows, HJKL and WASD...)
> 
>     3. Use whatever letters that fit and we can work on another PR to bring these letters to the config file 💡
> 
> 
> IMO it would bring a lot of value to start with just these few shortcuts. 🙇

Personally, I'd be happy to leave configurability for a followup feature. I could also live with just removing the number shortcuts (there's more than one reason against them), and then just mark all the rest as "Experimental" in the readme, which means we could still change them at a later point in time as we see fit.

But that's not even the biggest problem IMO.

- does IM Context introduced in #52 still work? Question asked [here](https://github.com/gabm/Satty/pull/148#issuecomment-2829903106), not answered. Somebody would have to test this.
- does anyone want to continue to work on this? Because there are conflicts that need resolving.

Issue #106, opened 2024-07-21
Add UI element to change font size


diniamo:
The title is self-explanatory, the default font size is way too big for small images.

jacekpoz:
any suggestions on where the UI for that could be and what it could look like?

diniamo:
Probably an input field with +- buttons on the bottom, next to the SML buttons.

gabm:
I agree.. but i don't have a 100% good idea on how to keep the interface lean with the increasing number of options/settings

gabm:
there are some ideas in #42 and #49

00-kat:
The SML buttons could be replaced with a
`- A +`
when the font tool is selected maybe. The A could be the current font size, instead of showing the number.

I was thinking of maybe sending a PR which replaces SML with - o + where o is the current pen size (after I figure out GTK), so if I ever do that this would fit in even more...

diniamo:
Could work to be honest.

RobertMueller2:
I've had a brief look here. My personal favourite would be to expose annotation-size-factor with 2 digits via a SpinButton in order to make this configurable at run time. It's not a font size, but if you set this to, e.g. 0.25, it's reasonably small. Given how the sizes are currently implemented, I think that is the least invasive approach.

I'm happy to try it, if anyone can help me with a hint where to store this runtime value for later access.

The toolbar side is flexible, it fires an event and can access app_config or style or whatever. But reading it in size impl is the tricky part, because size currently is an enum. that leaves two options IMO:
- overwrite this in configuration (and perhaps add default-annotation-size-factor so it can also be restored to config/cli value easily with a reset button). I tried this but ended up with deadlock from the control, probably can be solved. But since nothing else is doing it like that, I wasn't sure if this really is the way to go
- change Size so it knows whatever object stores the runtime value or convert Size into a struct that holds it. Also didn't really feel right.

Any thoughts on this?

PR #154, opened 2025-04-06
make annotation_size_factor configurable from UI

RobertMueller2:
Related to #106 

satty-20250406-090844
(screenshot outdated, see https://github.com/gabm/Satty/pull/154#issuecomment-2798812396)

- adds UI element to adjust `annotation_size_factor` config setting, using GTK SpinButton control. This affects the sizes of all tools and allows for finer control of the annotation sizes ;)
  - the SpinButton control offers by GTK default
    - left click to increase/decrease
    - right click to go to min/max
    - middle click to increase/decrease by a bigger jump
- adds reset  button allowing to going back to the default/configured value

The original goal of the issue was to make the font/font size configurable. I think somebody can implement that later, for now this seemed like a much quicker win.

TODO Items:
-  [x] make reset button more visible, not sure yet what the issue is, according to `GTK_DEBUG=interactive` it has the same color as the others. Edit: it was "edit-reset" vs "edit-reset-symbolic"
- [x] consider GTK drawer for the additional UI elements
- [x] allow focus of the text input box. i disabled this for now as it stole focus from the text annotation tool - no longer a concern because the control has been moved into a dialog

Also, I'm not really familiar with Relm. I couldn't say if all of this is the best possible way to implement it. It does work here and didn't seem to break anything. ;) I have tried adjusting the `APP_CONFIG` with `write` but that led to deadlocks (or `WouldBlock` with `try_write`) -- since nothing else writes except for intially command line parameters and config, this might not have been the way to go anyway.

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -363,6 +363,12 @@ impl SketchBoard {
                     .borrow_mut()
                     .handle_event(ToolEvent::StyleChanged(self.style))
             }
+            ToolbarEvent::AnnotationSizeChanged(value) => {
+                self.style.annotation_size_factor = value;
+                self.active_tool
+                    .borrow_mut()
+                    .handle_event(ToolEvent::StyleChanged(self.style))
+            }
         }
     }
 }
--- a/src/style.rs
+++ b/src/style.rs
@@ -11,11 +11,12 @@ use relm4::gtk::gdk::RGBA;
 
 use crate::configuration::APP_CONFIG;
 
-#[derive(Clone, Copy, Debug, Default)]
+#[derive(Clone, Copy, Debug)]
 pub struct Style {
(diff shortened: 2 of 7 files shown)


RobertMueller2:
Revealing/hiding the elements now looks like this:

satty-20250407-081608
satty-20250407-081657

RobertMueller2:
dealing with the focus in a consistent way, considering tab, enter, potentially escape, cursor up/down and pgup/pgdown is trickier than I thought. Even using the hide button as a possible focus target doesn't make it much better. 

Either the text box of the spin button remains unfocusable, or this needs to go in a dialog.

RobertMueller2:
Need to draft this again, this broke the color dialog.

gabm:
thanks, looks great

PR #155, opened 2025-04-22
feat: add option to allow copy to clipboard by right click

DreamMaoMao:
After editing the picture with the mouse, if you can directly use the right-click button to copy, it will greatly increase the copying speed, because there is no need to carefully position the mouse on the copy button, nor is there a need to touch the keyboard separately

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
 action-on-enter = "save-to-clipboard"
 # After copying the screenshot, save it to a file as well
 save-after-copy = false
+# Right click to copy
+right-click-copy = false
 # Hide toolbars by default
 default-hide-toolbars = false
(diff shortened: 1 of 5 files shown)


gabm:
@DreamMaoMao linting fails

DreamMaoMao:
> @DreamMaoMao linting fails

Has been reformatted

RobertMueller2:
With this change, when I use tools such as numbered marker or text, the position where the elements are actually placed are off.

I'm not sure yet what the logic change here was that causes it.

DreamMaoMao:
> placed are off

what is the mean about 'placed are off' ?

RobertMueller2:
Yeah, I thought it happened always, but that wasn't correct. It only occurs after resizing the window really. So probably best to add a recording. ;)

This is before the commit:
recording_1600f3

this is after:
recording_main

Do you observe this after resizing as well?

DreamMaoMao:
> Do you observe this after resizing as well?

my bad, i forgot to update coordinates.

fix in:
https://github.com/gabm/Satty/pull/162

Issue #91, opened 2024-06-06
[bug] using the chrono formatting options for file name causes satty to hang and crash on save


a-viv-a:
When using the chrono formatting options, satty hangs and crashes. I used `--output-filename /home/aviva/screenshots/%Y-%b/%a-%H:%M:%S:%1f.png` (probably change to be appropriate on your system?) and when I click the save button, satty locks up. If you wait long enough it seems to die, but the following trace is from when I `kill`ed satty.

```
thread 'main' panicked at library/alloc/src/fmt.rs:612:32:
a formatting trait implementation returned an error: Error
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: alloc::fmt::format::format_inner
   4: satty::sketch_board::SketchBoard::handle_save
   5: relm4::component::sync::traits::Component::update_with_view
   6: glib::main_context_futures::<impl glib::auto::main_context::MainContext>::spawn_local_with_priority::{{closure}}
   7: glib::main_context::<impl glib::auto::main_context::MainContext>::with_thread_default
   8: glib::main_context_futures::TaskSource::dispatch
   9: g_main_dispatch
  10: g_main_context_iterate_unlocked.isra.0
  11: g_main_context_iteration
  12: g_application_run
  13: gio::application::ApplicationExtManual::run_with_args
  14: relm4::app::RelmApp<M>::run
  15: satty::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

same deal but "full" backtrace:

```
thread 'main' panicked at library/alloc/src/fmt.rs:612:32:
a formatting trait implementation returned an error: Error
stack backtrace:
   0:     0x55ffd0b781f7 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h83d3e31f1809523c
   1:     0x55ffd0bacf00 - core::fmt::write::h76573a5842288faa
   2:     0x55ffd0b92255 - std::io::Write::write_fmt::hb095d534967ab243
   3:     0x55ffd0b77fc4 - std::sys_common::backtrace::print::h73765b495122298a
   4:     0x55ffd0b78b37 - std::panicking::default_hook::{{closure}}::hfad133b7877f7d1a
   5:     0x55ffd0b78897 - std::panicking::default_hook::hb7ccbed45ddd154f
[...]

PR #161, opened 2025-05-01
Issue #91: intercept chrono panic

RobertMueller2:
When DelayedFormat is used, and the format string cannot be processed (e.g. invalid), this causes a panic.

See related discussion in
https://github.com/chronotope/chrono/issues/1588

Use fallible function in the future once available, for now: catch_unwind the panic. 

I'm not auto closing the issue #91 from here, we might still need that for tracking purposes for future changes here.

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -193,7 +193,7 @@ impl SketchBoard {
     }
 
     fn handle_save(&self, image: Pixbuf) {
-        let output_filename = match APP_CONFIG.read().output_filename() {
+        let mut output_filename = match APP_CONFIG.read().output_filename() {
             None => {
                 println!("No Output filename specified!");
                 return;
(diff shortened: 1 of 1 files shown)


gabm:
looks great.. I'd merge it. Just one small question: do you think that we could "just do something senseful" anyway? i.e. convert your "error" to a "warning" and save the file?

RobertMueller2:
hmm, so just use the filename literally? Yeah, I suppose that works.

RobertMueller2:
Falling back to literal filename now. I mean, of course that could overwrite an existing file in worst case, I hope users would notice long before that.

Alternatively we could have replaced and randomised the file basename. Not a biggie but seems a bit much for this very specific error situation. Happy to do that later if we ever get complaints about overwritten literal filenames ;)

PR #162, opened 2025-05-01
fix: miss update coordinates in mouse  click event

DreamMaoMao:
fix the bug from the pr:
https://github.com/gabm/Satty/pull/155


RobertMueller2:
lgtm, thanks for fixing this.

PR #165, opened 2025-05-02
add alt modifier to ellipse tool

### LINKED ISSUE #125: Draw circle from the corner instead of the center?
In most graphics and drawing apps, circles/ellipses are usually drawn from the corner, not the center; usually a modifier (ctrl or shift i think usually) will expand from the center

What would it take to get the circle to start from the corner, like the rectangle? I find it's easier to circle the things I want that way (it's how swappy works, but i find satty a better app overall)

Thanks!

RobertMueller2:
Closes #125 

This allows dragging circle from corner.

Ctrl+shift draws a circle from corner.

circle_corner

I suppose we could think about swapping around control and not control.

### DIFF
--- a/src/tools/ellipse.rs
+++ b/src/tools/ellipse.rs
@@ -16,6 +16,8 @@ pub struct Ellipse {
     middle: Vec2D,
     radii: Option<Vec2D>,
     style: Style,
+    centered: bool,
+    finishing: bool,
 }
 
 impl Drawable for Ellipse {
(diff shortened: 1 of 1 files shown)


gabm:
looks great, that was a comment on reddit yesterday :)

I have two questions

- does this match the logic of other tools (default vs. shift vs. Ctrl)? two come to mind: PowerPoint and Excalidraw)
- what happens if you drop the modifier midway? e.g. start with Ctrl, let it go and then press it again?

RobertMueller2:
Re 1: yeah, I thought I might have to rework that. :smile:  I only have access to powerpoint online, but that has no modifier keys and defaults to "from corner" mode. Excalidraw defaults to "from corner', uses Alt to use "from center", and shift to make it a circle. Perhaps I can think of more other apps to check and we might see inconsistencies between alt and ctrl, but it kinda looks like "from corner" should be the default behaviour.

Re 2: since I'm storing the origin in the ellipse, I can start over with whatever the mode is in that very moment. E.g. start with ctrl, it draws "from corner", let go of ctrl, the circle is now drawn "from center" starting with the original origin. That behaviour matches Excalidraw.

gabm:
i don't think we need to check all the other tools.. from corner as default and shift for "circle" are default imho. Let's go with alt for center and that's it?!

RobertMueller2:
That's changed now, using none for "from corner" and alt for "from circle"
circle_corner2

I've added some helper indicators:
- a "bounds" rect for the "from corner" mode
- a center mark (circle) for the "from circle" mode, just while dragging.

Felt like a nice addition, but I can take it out again if it's too much. (Would have been nice to use a dashed line, but that seems overly complicated with femtovg). The hardcoding (color/paint width) isn't ideal, but can be revisited if we ever touch other tools.

gabm:
that looks great, thanks so much!

PR #167, opened 2025-05-08
add --profile-startup

RobertMueller2:
Parameter only but no config option (for now). This prints profiling for app startup, e.g.

```
startup timestamp was 1746683407.715279841 2025-50-08 07:50:07
    1 ms time elapsed: "configuration loaded"
    1 ms time elapsed: "loaded gl"
    1 ms time elapsed: "loading image"
  412 ms time elapsed: "image loaded, starting gui"
  474 ms time elapsed: "app init end"
  540 ms time elapsed: "gui show event"
  543 ms time elapsed: "main loop idle"
```

Useful to combine with LD_DEBUG, e.g.
`date "+%s,%N" && LD_DEBUG=statistics  satty --filename ~/Downloads/big_noise.png --profile-startup`

### DIFF
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -72,6 +72,10 @@ pub struct CommandLine {
     /// Disable notifications
     #[arg(long)]
     pub disable_notifications: bool,
+
+    /// print profiling
+    #[arg(long)]
+    pub profile_startup: bool,
 }
 
 #[derive(Debug, Clone, Copy, Default, ValueEnum)]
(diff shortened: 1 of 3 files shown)


gabm:
> EDIT: LazyLock requires at least rustc 1.80, is that a problem? E.g. Debian would only be able to provide that from testing, but not in stable. I could maybe add a feature which is enabled by default?

Lets see whether someone complains. On the long run we might remove that again and can then lower that requirement

Issue #170, opened 2025-05-08
Configuration file broken in 0.18.0


gabm:
Reported in #132 by @braun-steven

> I was going to run this, but apparently v0.18.0 doesn't work for me? 
> 
> ```bash
> # Downloaded from GitHub release
> ❯ ./satty --version && ./satty -f image.jpg
> satty 0.18.0
> Error reading config file: Error reading file: No such file or directory (os error 2)
> 
> # Built 0.18.0 from source
> ❯ ./target/release/satty --version && ./target/release/satty -f image.jpg
> satty 0.18.0
> Error reading config file: Error reading file: No such file or directory (os error 2)
> 
> # Built 0.17.0 from source
> ❯ ./target/release/satty --version && ./target/release/satty -f image.jpg
> satty 0.17.0
> # (Satty successfully opens)
> ```
> 
> I'm on Fedora 42
> 
> ```bash
> ❯ uname -a
> Linux tuxedo 6.14.4-300.fc42.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Apr 25 15:43:38 UTC 2025 x86_64 GNU/Linux
> ``` 

 _Originally posted by @braun-steven in [#132](https://github.com/gabm/Satty/issues/132#issuecomment-2862437769)_

gabm:
that might have something to do with the upgraded xdg dependency. Do you have a config file? if yes, where is it?

braun-steven:
No config file. Any way to output debug logs or verbose logs?

braun-steven:
Ah, adding a config file (default from the README) at the default location resolves the issue. Seems like the case when no config is present is broken right now.

gabm:
okay thanks.. yes, something is broken there.. we will have a look

Liu-WeiHu:
I created an empty file ~/.config/satty/config.toml and it works!

gabm:
the fix is released as 0.18.1

PR #174, opened 2025-05-08
feature: escape action

### LINKED ISSUE #172: Feature request / --action-on-{exit,escape}
## What I'd like

Like `--action-on-enter` being able to set an action (copy) on exit (or escape)

fabienjuif:
fixes #172

### How

I've reused the action on enter idea.
I've made 3 new actions but we could argue that `early_exit` is enough.

### Example

```
➜  void-packages git:(satty) ✗ grim -t png -g "$(slurp -d)" - | satty -f - --action-on-escape=save-to-clipboard-and-exit --copy-command=wl-copy
config file not found
Copied to clipboard.
➜  void-packages git:(satty) ✗ 
```

image

### Misc

My first attempt was roughly this but it was unconsistent:
```rust
    // /!\ not consistent
    pub fn request_render_then(&self, action: Action, callback: impl FnOnce() + 'static) {
        let handler_id_rc: Rc<RefCell<Option<glib::SignalHandlerId>>> = Rc::new(RefCell::new(None));
        let callback_rc: CallbackBox = Rc::new(RefCell::new(Some(Box::new(callback))));
    
        let obj = self.obj();
        let handler_id_clone = handler_id_rc.clone();
    
        let handler_id = obj.connect_render(move |gl_area, _| {
            // call the callback
            if let Some(callback) = callback_rc.borrow_mut().take() {
                callback();
            }
            // disconnect handler
            if let Some(id) = handler_id_clone.borrow_mut().take() {
                gl_area.disconnect(id);
            }
            glib::Propagation::Proceed
        });
    
        handler_id_rc.borrow_mut().replace(handler_id);
        self.request_render.borrow_mut().replace(action);
        obj.queue_render();
    }
```

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -185,12 +185,26 @@ impl SketchBoard {
 
     fn handle_render_result(&self, image: RenderedImage, action: Action) {
         match action {
-            Action::SaveToClipboard => self.handle_copy_clipboard(Self::image_to_pixbuf(image)),
-            Action::SaveToFile => self.handle_save(Self::image_to_pixbuf(image)),
+            Action::SaveToClipboard | Action::SaveToClipboardAndExit => {
+                self.handle_copy_clipboard(Self::image_to_pixbuf(image))
+            }
+            Action::SaveToFile | Action::SaveToFileAndExit => {
+                self.handle_save(Self::image_to_pixbuf(image))
+            }
+            _ => (),
         };
-        if APP_CONFIG.read().early_exit() {
-            relm4::main_application().quit();
+        if APP_CONFIG.read().early_exit()
+            || matches!(
+                action,
+                Action::Exit | Action::SaveToFileAndExit | Action::SaveToClipboardAndExit
+            )
+        {
+            self.handle_exit();
         }
+    }
+
+    fn handle_exit(&self) {
+        relm4::main_application().quit();
     }
 
     fn handle_save(&self, image: Pixbuf) {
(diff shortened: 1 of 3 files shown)


RobertMueller2:
Thanks for the PR. I'll review this in a bit. Sorry for the additional trouble, but would you mind using the "self review" commit as a fixup commit for the first, and openening a separate PR for the fix for #170?

fabienjuif:
I am doing it.

I am curious, why do not you use "squash merge"?

src/sketch_board.rs
```diff
@@ -489,9 +494,17 @@ impl Component for SketchBoard {
...
-                        ToolUpdateResult::Unmodified
+                        // First, let the tool handle the event. If the tool does nothing, we can do our thing (otherwise require a second Escape)
+                        // Relying on ToolUpdateResult::Unmodified is probably not a good idea, but it's the only way at the moment. See discussion in #144
+                        let result: ToolUpdateResult = self
+                            .active_tool
+                            .borrow_mut()
+                            .handle_event(ToolEvent::Input(ie));
+                        if let ToolUpdateResult::Unmodified = result {
+                            self.renderer
+                                .request_render(APP_CONFIG.read().action_on_escape());
```
RobertMueller2, line 505: This is the only different line from the next section, isn't it?

To reduce code duplication, may I suggest joining the two sections like
```
} else if ke.key == Key::Return
                        || ke.key == Key::KP_Enter
                        || ke.key == Key::Escape
```

and then something like
```
self.renderer.request_render(if ke.key == Key::Escape {
                                APP_CONFIG.read().action_on_escape()
                            } else {
                                APP_CONFIG.read().action_on_enter()
});
```
?

I don't see why we would have to separate them again anytime soon. Let that be a future problem ;)

RobertMueller2:
> I am doing it.
> 
> I am curious, why do not you use "squash merge"?

Well, it's a valid question. I suppose in such an easy and obvious case we could. But generally spoken, the author may have had reasons to create separate commits which should also be visible in commit history after merging. For bigger PRs it could become difficult to distinguish this from mere formatting changes. So with that in mind, personally, I'd rather prefer squashing that in the PR. But it's no more than a personal feeling. I don't know how @gabm feels about this. ;)

RobertMueller2 (changes requested):
I think this could do with another `cargo fmt`, right? Other than that and the comment above,  looks good to me.

fabienjuif:
image

`grim -t png -g "$(slurp -d)" - | cargo run -- -f - --action-on-escape save-to-clipboard-and-exit --copy-command=wl-copy`

RobertMueller2:
github doesn't let me right now, I'll try again later.

PR #175, opened 2025-05-08
Makefile default fix fmt and clippy

fabienjuif:
I find it useful, up to you!


RobertMueller2:
hmm.. don't mind, but really change the default? Could it be a rule further down?

fabienjuif:
Modified @RobertMueller2

Issue #171, opened 2025-05-08
Feature request / Scroll on annotation size to change value


fabienjuif:
## What I'd like

When I scroll in this area, I'd like the value to change without having to open the modal.

![Image](https://github.com/user-attachments/assets/aff9d426-4226-4d32-ae8d-ccf4014ede38)

Issue #51, opened 2024-02-20
Slider Bar for Blur-intensity, Brush-width, Text-size, anything else which effects can be amplified


KAGEYAM4:
Currently these is predefined `Small, Medium, Large` for selecting width of drawn elements. It would be more easies if these was a single seek-bar which can be used to increase effect of current tool.

gabm:
I also thought about something like this. Currently we have a problem w.r.t. scaling: we don't know, which DPI the original image has/had and therefore we don't really know, what "small", "medium" and "large" should be. 

That being said, I couldn't work out a proper UI/UX concept for a slider.. I came up with something like this, reactive to mouse wheel for "up/down":

![image](https://github.com/gabm/Satty/assets/482105/aa4651eb-46c0-4107-8dbe-923e290e0d50)


What I don't like about this is that we lose the simplicity. I really wanted to avoid displaying numeric values

KAGEYAM4:
> What I don't like about this is that we lose the simplicity. I really wanted to avoid displaying numeric values

I was thinking something like this 
![image](https://github.com/gabm/Satty/assets/75798544/98a9b7df-f30e-4dfa-9145-f4668f35195d)


Maybe someone else has better ideas, lets wait.

KAGEYAM4:
What do you think about this, RightClick-hold-Drag up/down | left/right to Decrease/Increase intesity/width/size etc.

the main idea is that, user have to think less about where tool is and clicking on that value to get idea about the size, and can just use right-clik-hold-slide to do it more efficietly.

rtgiskard:
I think replace the <S M L> buttons with a scaler would be fine, and associate it with mouse scroll up/down to change the value, that would be simple and convenient enough

PR #178, opened 2025-05-08
S,M,L connected to SpinButton

fabienjuif:
related to #171

image

Idea is to simplify the UX a bit.
- The sizes are percentages
- We could add back button with a "size" that translate this way (this is what's exposed in the screenshot)
  - S(mall) => 33%
  - M(edium) => 100%
  - L(arge) => 300%

For now the PR breaks things, like
- Text inputs
  - Because the focus in always in the SpinButton atm

I find it more intuitive like this but this is maybe me?
WDYT?

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -16,7 +16,6 @@ pub struct Style {
     pub color: Color,
     pub size: Size,
     pub fill: bool,
-    pub annotation_size_factor: f32,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
@@ -27,21 +26,19 @@ pub struct Color {
     pub a: u8,
 }
 
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Default)]
-pub enum Size {
-    Small = 0,
-    #[default]
-    Medium = 1,
-    Large = 2,
+#[derive(Debug, PartialEq, PartialOrd, Clone, Copy)]
+pub struct Size {
+    pub value: f32,
 }
 
 impl Default for Style {
     fn default() -> Self {
         Self {
             color: Color::default(),
-            size: Size::default(),
             fill: bool::default(),
-            annotation_size_factor: APP_CONFIG.read().annotation_size_factor(),
+            size: Size {
+                value: APP_CONFIG.read().annotation_size_factor(),
+            },
         }
     }
 }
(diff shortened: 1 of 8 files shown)


fabienjuif:
image

fabienjuif:
related #51

fabienjuif:
WDYT @gabm  @RobertMueller2 ?

I think it way more usable like this but I am biased ofc

fabienjuif:
I think this is what I suggest here (or at least try)

> We could add back button with a "size" that translate this way (this is what's exposed in the screenshot)
>
>    S(mall) => 33%
>    M(edium) => 100%
>    L(arge) => 300%


There are not random values, this is values I tried in this PR and showed in the first screenshot

RobertMueller2:
ok, sorry, I suppose I got confused. Does indeed sound like the same thought.

fabienjuif:
image

With latest commit
I am unsure how to update the toolbar model on the action event yet

src/style.rs
```diff
@@ -165,81 +162,34 @@ impl From<Style> for Paint {
...
 }
 
 impl Size {
-    pub fn to_text_size(self, size_factor: f32) -> i32 {
-        match self {
-            Size::Small => (36.0 * size_factor) as i32,
-            Size::Medium => (54.0 * size_factor) as i32,
-            Size::Large => (96.0 * size_factor) as i32,
-        }
+    pub fn to_text_size(self) -> f32 {
```
fabienjuif, line 172: At one point it should be resolved to each tools maybe

fabienjuif:
At this stage I have s,m,l group buttons that modify the spin value.
Before opening the PR, what I'd like is when I change the spin value for something <> than the s,m,l no button is toggled on

gabm:
hmm gnome guidelines say: 

> Only use a spin button when the exact numeric value is meaningful or useful. If this isn’t the case, a slider might be a better choice.

https://developer.gnome.org/hig/patterns/controls/spin-buttons.html

maybe that's a better choice?!

fabienjuif:
Ok so we are going closer to the first version I've made without the S/M/L buttons.
And we could replace the spin with a slider.

If you feel this is worth a try I'll dig into that

RobertMueller2:
> Ok so we are going closer to the first version I've made without the S/M/L buttons. And we could replace the spin with a slider.
> 
> If you feel this is worth a try I'll dig into that

I think so. 
- S/M/L provides something that isn't too technical, disadvantage: lacks fine control
- Spin Button provides more control, but adds technicality

The solution on main does indeed solve the issue of lacking control, but it also combines the disadvantages of both.

I think the slider is perfect here, really, fine control if done right, doesn't expose the technical stuff.

MangelMaxime:
When reporting #243, someone pointed me to this PR

I tested the behavior of numbered marker in this PR and they are really too big (even bigger than before) 😅

image

From top to bottom, I used `L`, `M`, `S`

And I think this problem is the same for other element, like if you try to create an array with `L` size I don't know in which scenario it could be practical.

image

When comparing Satty to previous screenshot editor tool I used, I find that the sizing are not really that useful and often off for example to get a good sized Numbered Marker I would need to use size `S` but then arrows, rectangle, etc. feels too small and I need to choose size `M` or change the scale factor. This makes it difficult to quickly edit a screenshot

fabienjuif:
This PR aims also at simplifying our ratios for each tools.
We would needs, as you point it out, to tweak them before merging this PR

edit:

If you play with this PR and find tool factor that you think are best, it would be cool to share this so I can try them out too!

w-jablonski:
Numbered marker is something else than the other tools, which are line-based. It might be hard to find its default size (even when related to the line size or image size) that users would find satisfactory/non-surprising. So the user puts the numbered marker kinda "hit or miss" and will sometimes be dissatisfied with the size. So they hit Undo, resize with the slider or otherwise, and try again. And if still not satisfied, try again... Then once they've found the right size, now the size of other tools might be off... 

Perhaps an alternative would be to have numbered markers resizable as they appear, in similar manner as the "Ellipse tool" is resizable with Shift+Alt held. Then the next marker by default keeps the size of the original one. I think splitting the sizing of `numbered markers | line | text` would benefit UX. Even better, this particular one is a non-setting (the user chooses, not us).

 Thanks

RobertMueller2:
> Numbered marker is something else than the other tools, which are line-based. It might be hard to find its default size (even when related to the line size or image size) that users would find satisfactory/non-surprising. So the user puts the numbered marker kinda "hit or miss" and will sometimes be dissatisfied with the size. So they hit Undo, resize with the slider or otherwise, and try again. And if still not satisfied, try again... Then once they've found the right size, now the size of other tools might be off...
> 
> Perhaps an alternative would be to have numbered markers resizable as they appear, in similar manner as the "Ellipse tool" is resizable with Shift+Alt held. Then the next marker by default keeps the size of the original one. I think splitting the sizing of `numbered markers | line | text` would benefit UX. Even better, this particular one is a non-setting (the user chooses, not us).
> 
> Thanks

Hmm.. I can see some benefit in setting text and line width separately. But since the markers display text as well, surely that could be one setting for both, no?

Also, I'd say that resizable markers is not a thing in the other annotation tools I've used. I mean, sure, that's not a good reasoning at all. :D But I'd like to avoid creating something that is equally frustrating ;)

PR #179, opened 2025-05-08
save to io::stdout (aliased to :"-")

### LINKED ISSUE #126: Output to stdout
print to stdout if **-o --output-filename** is: **-**
cool feature to create own pipeline
_swappy_ do the same thing

fabienjuif:
fixes #126

### DIFF
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -15,7 +15,7 @@ pub struct CommandLine {
     #[arg(long)]
     pub fullscreen: bool,
 
-    /// Filename to use for saving action. Omit to disable saving to file. Might contain format
+    /// Filename to use for saving action or '-' to print to stdout. Omit to disable saving to file. Might contain format
     /// specifiers: <https://docs.rs/chrono/latest/chrono/format/strftime/index.html>.
     #[arg(short, long)]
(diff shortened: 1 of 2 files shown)


fabienjuif:
image

RobertMueller2:
Thanks for that. :)

Hmm.. I think allowing `/dev/stdout` (and perhaps more device nodes at some later point, not now, perhaps with a check whether it's a device node) is a good idea.

When we explicitly get `-`, I'd prefer to use io::stdout rather than aliasing it to `/dev/stdout`, even if it adds an additional if/else and thus makes the code a little bit uglier. Reason being, while `/dev/stdout` is probably available everywhere where anyone's using satty right now, it's not available everywhere. It's not Posix, and it's not on Windows. And while I don't think it even builds on Windows at this point, I'd rather try and not add additional obstacles when we can avoid them. Make sense?

fabienjuif:
image

`grim -t png -g "$(slurp -d)" - | cargo run -- -f - -o - --action-on-escape=save-to-file-and-exit | cargo run -- -f - --action-on-escape=save-to-clipboard-and-exit --copy-command=wl-copy`

PR #180, opened 2025-05-08
reset all modifications

### LINKED ISSUE #166: feature: clear all annotations
As far as I can tell, there's only `undo/redo` but no option to clear everything at once. Would be nice to have this function, ideally tied to a keyboard shortcut.

fabienjuif:
fixes #166

Also I've seen a bug in the redo/undo, it feels like there is "ghosts" actions in the stack.
If you think I am not hallucinating, I'll open a PR.

### screenshots
**icon**
image

image

**clicking 4 times on undo**
image

**redo all**
image

**reset**
image

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -345,6 +345,16 @@ impl SketchBoard {
         }
     }
 
+    fn handle_reset(&mut self) -> ToolUpdateResult {
+        if self.active_tool.borrow().active() {
+            self.active_tool.borrow_mut().handle_undo()
+        } else if self.renderer.reset() {
+            ToolUpdateResult::Redraw
+        } else {
+            ToolUpdateResult::Unmodified
+        }
+    }
+
     // Toolbars = Tools Toolbar + Style Toolbar
     fn handle_toggle_toolbars_display(
         &mut self,
@@ -413,6 +423,7 @@ impl SketchBoard {
             }
             ToolbarEvent::Undo => self.handle_undo(),
             ToolbarEvent::Redo => self.handle_redo(),
+            ToolbarEvent::Reset => self.handle_reset(),
             ToolbarEvent::ToggleFill => {
                 self.style.fill = !self.style.fill;
                 self.active_tool
--- a/icons.toml
+++ b/icons.toml
@@ -8,6 +8,7 @@ icons = [
   "highlight-regular",
   "arrow-redo-filled",
   "arrow-undo-filled",
+  "recycling-bin",
   "save-regular",
   "copy-regular",
   "text-case-title-regular",
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -269,6 +269,19 @@ impl FemtoVgAreaMut {
             None => false,
         }
     }
+    pub fn reset(&mut self) -> bool {
+        let mut any_undone = false;
+        while let Some(mut d) = self.drawables.pop() {
+            // notify of the undo action
+            d.handle_undo();
+
+            // push to redo stack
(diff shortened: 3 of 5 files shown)


src/sketch_board.rs
```diff
@@ -336,6 +336,16 @@ impl SketchBoard {
         }
     }
 
+    fn handle_reset(&mut self) -> ToolUpdateResult {
+        if self.active_tool.borrow().active() {
+            self.active_tool.borrow_mut().handle_undo()
```
fabienjuif, line 341: Not sure about this

gabm: that was introduced so that the text tool can handle undo/redo itself

gabm:
I think the functionality is nice, but I have two requests

- can we use another icon? typically its an eraser...
- maybe the icon should be third after undo/redo?!

gabm:
I dont know about the bug you describe, but if its real then lets create an issue

RobertMueller2:
> I think the functionality is nice, but I have two requests
> 
>     * can we use another icon? typically its an eraser...
> 
>     * maybe the icon should be third after undo/redo?!

considering that "undo all" is different from a future "delete all", and there could be a "redo all" as well, I think I'd like
- a button that indicates "dramatic undo"
- or just a right click on the undo button that is reflected in tooltip and README?

fabienjuif:
> or just a right click on the undo button that is reflected in tooltip and README?

I feel like the discoverability is poor.
But I also understand the idea of not bloating the UI.
Maybe long press? A bit like firefox but without the history?
Long press -> drop down -> clear all?

I dunno.

---

I am open to change the icon, the order, etc.
Just tell me when you agree on a solution! :D

RobertMueller2:
> > or just a right click on the undo button that is reflected in tooltip and README?
> 
> I feel like the discoverability is poor. But I also understand the idea of not bloating the UI. Maybe long press? A bit like firefox but without the history? Long press -> drop down -> clear all?

Yup, I totally agree, it's poor that way. I think in the long run, buttons may have more than one function. Along the lines of how  Gimp does it. But it's probably something we have to discuss elsewhere and later, and carefully weight visibility/accessibility against UI clutter.

For now, my best suggestion is:
duplicate undo/redo buttons, color them differently (via style? if that is possible at all), and then they can represent "undo all" and "redo all"

How does that sound? @fabienjuif @gabm

fabienjuif:
I would rather find new icons than playing with colors.
People will dislike colors of buttons, while rn this is simple.

For the overall idea, agreed

fabienjuif:
image

gabm:
hmmm i like it.. pragmatic and functional. Referring to: 

- the basket icon and not positioning it as "undo all" but rather "delete all"
- the pop-up menu for lesser used options
- the top-and-left toolbar instead of top-and-bottom.. reduces the distance for the mouse to travel

RobertMueller2:
Well, if there's a basket icon, let's use that for now.  If we ever add another delete method, we can revisit later.

fabienjuif:
image

RobertMueller2:
hmm... I'm not sure about the exact repro, but definitely related to text. With random clicking I somehow caused the undo stack to be almost empty, while there still were a lot of elements on the screen.

redo

AFAICT we already have that problem without the reset button. Could be the "not sure about this" part. ;)

fabienjuif:
I am wondering if this is not the tool.undo()/redo().
But like you said this is not related to this PR, that we are hijacking :P

RobertMueller2:
so we just create a new issue for this?

fabienjuif:
Yep I am up for this

PR #182, opened 2025-05-08
brush input smoothing

### LINKED ISSUE #94: Input smoothing / post-processing
It would be nice if satty could implement mouse input post-processing (e.g. smoothing) similar to the way Windows snipping tool "prettifies" the mouse-drawn input.

Compare (deliberately sloppy) attempts to draw an ellipse with a laptop touchpad with Satty and with Windows 10 snipping tool (<kbd>Win</kbd> + <kbd>Shift</kbd> + <kbd>S</kbd>):

Satty:
image

Windows 10 snipping tool:
image

---

Note that this Windows 10 instance is running in a VM using host input, so there should be no difference due to e.g. touchpad drivers — the Windows app received exactly same input events as a Linux application would have in its place.

fabienjuif:
fixes #94 

with custom implem

### DIFF
--- a/src/command_line.rs
+++ b/src/command_line.rs
@@ -84,6 +84,12 @@ pub struct CommandLine {
     /// Disable the window decoration (title bar, borders, etc.)
     #[arg(long)]
     pub no_window_decoration: bool,
+
+    /// How much points to use for the brush smoothing algorithm
+    /// 0 disables smoothing.
+    /// The default value is 0 (disabled).
+    #[arg(long)]
+    pub brush_smooth_history_size: Option<usize>,
 }
 
 #[derive(Debug, Clone, Copy, Default, ValueEnum)]
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -51,6 +51,7 @@ pub struct Configuration {
     disable_notifications: bool,
     profile_startup: bool,
     no_window_decoration: bool,
+    brush_smooth_history_size: usize,
 }
 
 #[derive(Default)]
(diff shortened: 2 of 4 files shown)


gabm:
nice! that would be a great addition - good find!

RobertMueller2:
Nice find and it looks good to me. But I can also notice this makes the brush tool slower, at first glance this seems to be related to image size. 10000x10000 may not be the typical use case, but still, once I've seen that I can notice a slight delay on lower sizes as well :smile: 

When using a mouse, I don't have the problem described in #94. So for this scenario I kinda lose some speed and gain nothing. On the other hand, the problem does clearly exist when using a touchpad or touchscreen. I can see a case for wanting to turn this off, perhaps even with a modifier and an option setting what the default should be?

What do you think?

fabienjuif:
I'll add the `brush_smooth_history_size` option.
**Defaulted to 5** (less than this PR)
And you can set it to 0 to disable it.

WDYT?

RobertMueller2:
Sounds good.

fabienjuif:
I would suggest to test a big image like this on main branch @RobertMueller2 .

RobertMueller2:
looks good now.

RobertMueller2:
@gabm (and @fabienjuif ) the paint is indeed a bit behind the cursor if it's activated at all. I think since we can disable the smoothing, and that works well with brush-smooth-history-size 0, this is good enough for v1.

If I had to guess (I hate guessing, but I need to take a shortcut here), I'd say other tools paint first and then correct a bigger set of points of the previous data. But I might totally barking up the wrong tree here, I have not spent a lot of effort to understand this at all, and I won't pretend that I understand it. ;)

My suggestion: let @fabienjuif say if he wants to try and improve this even more now before merging. Otherwise, I'm fine with merging this (after a rebase ofc) as v1/experimental and add further improvements to the algorithm later.

fabienjuif:
I am up to disable it by default AND merge it with the new option.
I'll try to clean that up this evening

gabm:
great decision, we can enable when required

PR #186, opened 2025-05-09
blur without black rectangle

### LINKED ISSUE #183: Remove the black outline around the blurred area
When I use the blur tool in Satty to blur a part of the image, it also adds a box around the blurred area with black outlines. Please see the image attached below. I have marked the black outlines with red arrows.

Image

Is this behaviour intentional? Would it be possible to add a configuration option to disable the black outlines?

Other relevant information:
I am running the latest release of Satty on Fedora 41 using the prebuilt binary provided on the [Satty Releases](https://github.com/gabm/satty/releases) page.

### LINKED ISSUE #185: Blurring outside of crop darkens the image
On `satty 0.18.0`

Blurring an area outside of the currently cropped area will darken the selected area on top of blurring it. Example:

https://github.com/user-attachments/assets/73f6af19-d2fe-406b-afd6-a38803e08bb5

fabienjuif:
fixes #183 
fixes #185

### DIFF
--- a/src/tools/blur.rs
+++ b/src/tools/blur.rs
@@ -92,6 +92,9 @@ impl Drawable for Blur {
             // draw
             canvas.stroke_path(&path, &paint);
         } else {
+            canvas.save();
+            canvas.flush();
+
             let (pos, size) = math::rect_ensure_positive_size(self.top_left, size);
 
             // create new cached image
@@ -127,6 +130,7 @@ impl Drawable for Blur {
                     1f32,
                 ),
             );
+            canvas.restore();
         }
         Ok(())
     }


src/tools/blur.rs
```diff
@@ -14,23 +14,42 @@ use crate::{
...
+
+    fn reset(&mut self, top_left: Vec2D) {
+        self.top_left = Some(top_left);
+        self.size = None;
+        self.editing = true;
+        self.cached_image.replace(None);
+    }
+
     fn blur(
+        &self,
```
fabienjuif, line 43: I was surprised it was not used like this.
If this is ok I'll remove pos/size and compute it IN blur (ie: not from the caller)

fabienjuif, line 43: On top of that (new conversation to have a new thread), I've seen the app crashes if we are taking an area that is bigger than the based image, so we could clamp the selected rectangle in this blur function too.

RobertMueller2: yup, can confirm that.

fabienjuif: https://github.com/gabm/Satty/issues/189

RobertMueller2:
lgtm, but I'll create a followup issue related to blur but not your change ;)

PR #188, opened 2025-05-10
triggers - actions

### LINKED ISSUE #187: Triggers into Vec<Actions>
Conversation from here: https://github.com/gabm/Satty/issues/152

The idea would be having these args:
```
--action-on-enter
--action-on-escape
--action-on-right-click
--action-on-copy
```

That perform a list of actions
```
SaveToClipboard,
SaveToFile,
Exit,
```

Example: `satty -f - --action-on-escape=save-to-clipboard,exit`

fabienjuif:
fixes #187 

## How to test
```
grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - 
--copy-command=wl-copy --action-on-escape=save-to-clipboard,exit
```

```
grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - 
--copy-command=wl-copy --action-on-escape=save-to-clipboard
```

## TODOs
- [x] Connect everything
- [x] Tests the other scenarii
- [x] Clippy
- [x] Doc (README.md)
  - [x] cli
  - [x] toml
- [x] Better clap --help
  - [x] Deprecated args
  - [x] Maybe add new args using plural forms, and deprecate the others?

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,17 +86,31 @@ copy-command = "wl-copy"
 annotation-size-factor = 2
 # Filename to use for saving action. Omit to disable saving to file. Might contain format specifiers: https://docs.rs/chrono/latest/chrono/format/strftime/index.html
 output-filename = "/tmp/test-%Y-%m-%d_%H:%M:%S.png"
-# Action to perform when the Enter key is pressed [possible values: save-to-clipboard, save-to-file]
-action-on-enter = "save-to-clipboard"
+
 # After copying the screenshot, save it to a file as well
 save-after-copy = false
-# Right click to copy
-right-click-copy = false
 # Hide toolbars by default
 default-hide-toolbars = false
 # The primary highlighter to use, the other is accessible by holding CTRL at the start of a highlight [possible values: block, freehand]
 primary-highlighter = "block"
+# Disable notifications
 disable-notifications = false
+# Actions to trigger on right click (order is important)
+# [possible values: save-to-clipboard, save-to-file, exit]
+actions-on-right-click = []
+# Actions to trigger on Enter key (order is important)
+# [possible values: save-to-clipboard, save-to-file, exit]
+actions-on-enter = ["save-to-clipboard"]
+# Actions to trigger on Escape key (order is important)
+# [possible values: save-to-clipboard, save-to-file, exit]
+actions-on-escape = ["exit"]
+
+# Action to perform when the Enter key is pressed [possible values: save-to-clipboard, save-to-file]
+# Deprecated: use actions-on-enter instead
+action-on-enter = "save-to-clipboard"
+# Right click to copy
+# Deprecated: use actions-on-right-click instead
+right-click-copy = false
 
 # Font to use for text annotations
 [font]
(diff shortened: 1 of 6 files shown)


fabienjuif:
What do you think of

>  Maybe add new args using plural forms, and deprecate the others?

@gabm @RobertMueller2 


---

Since we are 0.xx I think we could break retro-compatibility.
But you may not see this like it.
Maybe we should find a rule, like we keep compatibility for **n** versions? Or **n** days?

gabm:
could you make this a bit more explicit, i don't really understand nor the implications...

Ideally we only deprecate but never remove ..

fabienjuif:
Before we were able to set one action per _trigger_
Example: `--action-on-escape=save-to-clipboard`

Now we can set multiple actions (this is a vector, order is important):
Example: `--action-on-escape=save-to-clipboard,exit`

So in the code I renamed the fields to use `actionS` rather than `action` (without an s).

And I am wondering if we should do the same for the toml and the CLI.
And then I am wondering if we should remove the `--action-on-escape` (since we would have `--actions-on-escape`)

I understand we want a stable API, but also in the meantime we should not bloat the API.
So my thought was, since we are in a 0.x.y version, people should be aware that the API can still change.

gabm:
as far as I can see, the only offending option is `action-on-enter` introduced by @pol-rivero around half a year ago. `action-on-escape` has never been released yet, so we can change it without any difficulty.

My suggestion would be:

- introduce plural forms and make them default wherever sensible in config and cli
- do _not_ introduce any additional singular forms - remove `action-on-escape` as it has never been released yet
- deprecate `action-on-enter` and maybe remove in future .. maybe we could also hide it in clap - like this it doesn't do any harm and the extra code to handle it is minor imho

fabienjuif:
```
Usage: satty [OPTIONS] --filename <FILENAME>

Options:
  -c, --config <CONFIG>
          Path to the config file. Otherwise will be read from XDG_CONFIG_DIR/satty/config.toml
  -f, --filename <FILENAME>
          Path to input image or '-' to read from stdin
      --fullscreen
          Start Satty in fullscreen mode
  -o, --output-filename <OUTPUT_FILENAME>
          Filename to use for saving action or '-' to print to stdout. Omit to disable saving to file. Might contain format specifiers: <https://docs.rs/chrono/latest/chrono/format/strftime/index.html>
      --early-exit
          Exit directly after copy/save action
      --corner-roundness <CORNER_ROUNDNESS>
          Draw corners of rectangles round if the value is greater than 0 (Defaults to 12) (0 disables rounded corners)
      --initial-tool <TOOL>
          Select the tool on startup [aliases: init-tool] [possible values: pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, highlight, brush]
      --copy-command <COPY_COMMAND>
          Configure the command to be called on copy, for example `wl-copy`
      --annotation-size-factor <ANNOTATION_SIZE_FACTOR>
          Increase or decrease the size of the annotations
      --save-after-copy
          After copying the screenshot, save it to a file as well Preferably use the `action_on_copy` option instead
      --actions-on-enter <ACTIONS_ON_ENTER>
          Actions to perform when pressing Enter [possible values: save-to-clipboard, save-to-file, exit]
      --actions-on-escape <ACTIONS_ON_ESCAPE>
          Actions to perform when pressing Escape [possible values: save-to-clipboard, save-to-file, exit]
      --actions-on-right-click <ACTIONS_ON_RIGHT_CLICK>
          Actions to perform when hitting the copy Button [possible values: save-to-clipboard, save-to-file, exit]
  -d, --default-hide-toolbars
          Hide toolbars by default
      --font-family <FONT_FAMILY>
          Font family to use for text annotations
      --font-style <FONT_STYLE>
          Font style to use for text annotations
      --primary-highlighter <PRIMARY_HIGHLIGHTER>
          The primary highlighter to use, secondary is accessible with CTRL [possible values: block, freehand]
      --disable-notifications
          Disable notifications
      --profile-startup
          Print profiling
      --no-window-decoration
          Disable the window decoration (title bar, borders, etc.)
      --right-click-copy
          Right click to copy Preferably use the `action_on_right_click` option instead
      --action-on-enter <ACTION_ON_ENTER>
          Action to perform when pressing Enter Preferably use the `actions_on_enter` option instead [possible values: save-to-clipboard, save-to-file, exit]
  -h, --help
          Print help
  -V, --version
          Print version
➜  Satty git:(fix_187) 
```

fabienjuif:
right click

```
➜  Satty git:(fix_187) grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - --actions-on-right-click=exit
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/satty -f - --actions-on-right-click=exit`
config file not found
```
```
➜  Satty git:(fix_187) grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - --actions-on-right-click=save-to-clipboard
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/satty -f - --actions-on-right-click=save-to-clipboard`
config file not found
Copied to clipboard.
```

fabienjuif:
enter

➜  Satty git:(fix_187) grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - --actions-on-enter=save-to-clipboard,exit
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/satty -f - --actions-on-enter=save-to-clipboard,exit`
config file not found
Copied to clipboard.
```
➜  Satty git:(fix_187) grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - --action-on-enter=exit
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/satty -f - --action-on-enter=exit`
config file not found
➜  Satty git:(fix_187)

**Comment by fabienjuif** [comment]
escape
```
➜  Satty git:(fix_187) grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - --actions-on-escape=save-to-clipboard,exit
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/satty -f - --actions-on-escape=save-to-clipboard,exit`
config file not found
Copied to clipboard.
```

```
➜  Satty git:(fix_187) grim -t ppm -g "$(slurp -d)" - | cargo run  -- -f - --actions-on-escape=exit
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/satty -f - --actions-on-escape=exit`
config file not found
➜  Satty git:(fix_187) 
```

PR #191, opened 2025-05-10
ensure blur is in bounds

### LINKED ISSUE #189: crash when bluring outside viewport
## Reproduction

Try to blur after the window on the right (or bottom).
It'll crash the app.

fabienjuif:
fixes #189


RobertMueller2:
looks good to me.

PR #192, opened 2025-05-10
filled blur rectangle

### LINKED ISSUE #190: Blur drawing rectangle depends on annotation_size
When using blur, the area affected by it is indicated by a rectangle.

This rectangle depends on annotation_size. Personally I think that's not useful for this tool. E.g. when editing a large image 10000x10000 pixel and blur  at annotation_size_factor 0.25, the rectangle is not visible. (The same is true for the other tools, but for those it's what actually is drawn so makes sense.)

fabienjuif:
fixes #190

### DIFF
--- a/src/tools/blur.rs
+++ b/src/tools/blur.rs
@@ -9,7 +9,7 @@ use crate::{
     configuration::APP_CONFIG,
     math::{self, Vec2D},
     sketch_board::{MouseEventMsg, MouseEventType},
-    style::{Size, Style},
+    style::Style,
 };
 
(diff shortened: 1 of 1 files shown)


fabienjuif:
@RobertMueller2 what about having a semi transparent fill rather than a stroke? :thinking:

RobertMueller2:
> @RobertMueller2 what about having a semi transparent fill rather than a stroke? 🤔

That could work, might make it more visible even.

fabienjuif:
@RobertMueller2 I let you try?

RobertMueller2 (approved):

RobertMueller2:
Nice. Thanks!

PR #195, opened 2025-05-11
Update README.md

### LINKED ISSUE #74: [FEATURE] Resizing the output
Hi, first of all, thanks for this tool, it's really amazing!

The only thing that i missing for me is an ability to set a resize value in the config and/or in the command. I'm not 100% sure what would be the best approach so let me explain my workflow a little bit.

In my line of work, I use screenshots a lot to make documentation. I usually make documentations for normies and screenshots are really the way to go to help themnavigate through an app. When I write documentation, I use a template so all my documentation look alike and I'm mostly using a 2 columns document. The column on the left has a screenshot with arrows, annotations and labels and the column on the right has all the text describing what the user see and what he can do in the window showed on the left side. As you can expect, having a fixed sized image is a must in this kind of workflow.

I'm using Hyprland, loving it, and right now I've bind a couple of keys to pipe grimblast into Satty, it's working really well. What I'd like to do is to bind another key (Shift-Print in my case) to load a different config of Satty to resize the file to 300x300 pixels before saving the output.

First, we would obviously need to implement resizing in Satty and I'm not sure how much work that would be...

Second, I currently see 2 ways to implement this if we manage to get this in Satty

1. Have a new command `--resize` and set the number of pixels for the output. We could also implement it via % of initial file as I've seen in many places but in my case, I really need fixed resolutions to fit everything in my columns for the docs. This value could be set in the command line and could also be set in the config file. In this case, I would simply have 2 config files and I would call the right one with the right binding.

2. Set a `--doc` command to output both files, original and resized one in the chosen folder. This way we wouldn't have the need to have 2 separate config files.

I'm not used to Rust so I don't think I'll be of much help, but if you're in need, I'm always willing to learn, we just won't have this soon ;)

### LINKED ISSUE #181: Adjust README.md
I should have thought of that, but we need #174 in there. My bad, sorry.

While we're at it, add a remark why ppm is faster. While we use ppm in the example, we don't really say why.

And also make sure we've covered all modifiers.

RobertMueller2:
Closes #181 
Closes #74 

Some changes will have to go again very soon with #188.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,13 +60,22 @@ All configuration is done either at the config file in `XDG_CONFIG_DIR/.config/s
 
 ### Shortcuts
 
-- `Enter`: as configured (see below)
-- `Esc`: Exit application
+- `Enter`: as configured (see below), default: copy-to-clipboard
+- `Esc`: as configured (see below), default: exit
 - `Ctrl+C`: Save to clipboard
 - `Ctrl+S`: Save to specified output file
 - `Ctrl+T`: Toggle toolbars
 - `Ctrl+Y`: Redo
 - `Ctrl+Z`: Undo
+
+### Tool Modifiers
+
+- Arrow: Hold `Shift` to make arrow snap to 15° steps
+- Ellipse: Hold `Alt` to center the ellipse around origin, hold `Shift` for a circle
+- Highlight: Hold `Ctrl` to switch between block and freehand mode (default configurable, see below), hold Shift for a square
+- Line: Hold `Shift` to make line snap to 15° steps
+- Rectangle: Hold `Alt` to center the rectangle around origin, hold `Shift` for a square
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Back to draft: Will add something for #74 as well

gabm:
now that i think of it: the rectangle might also work as the ellipse tool with alt?

RobertMueller2:
> now that i think of it: the rectangle might also work as the ellipse tool with alt?

Sure, we could implement that.

fabienjuif (approved):
:+1:

fabienjuif (approved):
Thank you again! :heart:

fabienjuif:
If this was not on purposes, there are also actions now https://github.com/gabm/Satty/commit/243a170c5d43855caed07d466b277a5f36d37924

gabm:
thanks!

PR #200, opened 2025-05-12
idea / generate a conf file with --init-config

fabienjuif:


gabm:
yep nice addition.. In future we might have a GUI for editing though - but who knows..like it

Issue #31, opened 2023-12-06
Documentation


gabm:
With the increasing number of configuration options #30, we also need better documentation to explain them to the users. I want this to be available on the web-page (github, maybe a domain) in a searchable and readable form as well as a man page. Ideally we only have one source of truth for this.

As for the man pages, a lot of projects seem to use scd nowadays...

RobertMueller2:
This applies to script examples as well. We could place those in a subfolder in the medium term, I assume that can be automatically considered with static site generation.

Just leaving that comment here so we can merge #208 for now, and worry about cleaning up later.

RobertMueller2:
I've had a brief look. mkdocs 1.0 is deprecated, and mkdocs 2.0 doesn't support Material, apparently. I think we want to use Zensical, but I'm placing this on hold at least until Zensical reached phase 1b. https://zensical.org/compatibility/

Outline:
- avoid duplication of content
- `satty --help` feeds README AND documentation, and manpage via clap_mangen and #405 
- sections that are in the README and useful for the documentation are automatically reused in the documentation
- new documentation sections are generated from new sources
- perhaps move some parts from README to the documentation and/or examples subfolder
- auto generate documentation on push or merge to main via gh actions
- have versioned documentation upon each release
- have config file example point out defaults

PR #208, opened 2025-05-16
readme.md: sway mode example

fabienjuif:


RobertMueller2:
I'm merging this for now, but as we've noticed, we're slowly but surely cluttering up the main readme and need to think about documentation, splitting off stuff from the main readme, etc. See #31.

Issue #82, opened 2024-05-07
[feature] style: allow to load extra css file for customize


rtgiskard:
Just like other gtk utils, allow the load of extra css will be great

gabm:
okay.. do you have an example on how to do this?

rtgiskard:
> okay.. do you have an example on how to do this?

I'm not sure for the details, [ReGreet](https://github.com/rharish101/ReGreet) provide the ability to do that, and it's also written in rust with gtk.

RobertMueller2:
Supporting CSS with a GTK application wouldn't be difficult, although I suppose the individual GUI elements would need a lot more naming to make it useful.

But I'm not sure to what end, would you mind to explain the use case? Do you want to style the GUI elements of Satty?

rtgiskard:
Current looking with transparency seems to be fine, no need to add extra themes ;)

rubiin:
is this completed? I couldnt find any reference on the sample config file or readme

fabienjuif:
It is but not released yet I think

PR #212, opened 2025-05-17
min size and css customization

### LINKED ISSUE #173: Enhancement / Min window size == to tools size
## Issue

When I take a tiny part of my screen as screenshot all the tool are not available until I resize the container (I am using sway).

Image

## What I'd like

Having the window size printing all tools
Related to https://github.com/gabm/Satty/issues/93 I guess

fabienjuif:
related to #82
fixes #173

### DIFF
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -486,7 +486,7 @@ impl ConfigurationFile {
     }
 
     fn try_read_xdg() -> Result<Option<ConfigurationFile>, ConfigurationFileError> {
-        let dirs = BaseDirectories::with_prefix("satty");
+        let dirs = BaseDirectories::with_prefix(env!("CARGO_PKG_NAME"));
         match dirs.get_config_file("config.toml") {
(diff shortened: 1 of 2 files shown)


RobertMueller2:
hmm.. should we place the default CSS in an external file?

fabienjuif:
You can override it, try to set .root{} and it should remove everything from our default.
So I think this is better to have it embedded so this is lightweight.

RobertMueller2:
I didn't mean to expose the default, just have it in a different file in the repository, then use include_bytes macro. Since there was css before, this might not be needed for THIS particular PR, but maybe in the future we can move it out of there. Maybe if it grows any larger.

Github totally misses inline languages, btw ;)

fabienjuif:
Ok, can you do an issue from your comment maybe?
I'll try to do that I am interested!

RobertMueller2:
looks good to me.

PR #219, opened 2025-05-25
Update README.md

postsolar:
Clarify the effect of Shift modifier for Highlight tool


RobertMueller2:
Thanks!

I missed that when adding the section.

PR #221, opened 2025-05-25
fix: don't use actions-on-enter and actions-on-escape when not needed

### LINKED ISSUE #218: bug: exit on shift-enter when entering text if actions-on-enter includes exit
Reproducible with
```
grim -t ppm - | satty -f - -o - -c <(echo 'general.actions-on-enter = [ "exit" ]')
```

Select the text instrument and it would exit.

Satty version: 0.19.0 45741b69a8efc60bb735ec256df3c9d90d8edda6

postsolar:
Only use actions-on-enter and actions-on-escape if no modifier masks are applied and nothing is done by the active tool.

Closes #218


fabienjuif:
Thank you @postsolar

PR #228, opened 2025-05-28
Remove tokio from direct dependencies

oakrotka:
Tokio is only used in one place, where it can be replaced with a call to rust's standard library. While tokio itself is already a dependency od relm4, this project enables all of tokio's features, which causes several unnecessary libraries to be compiled with it.

Eventually, if I'm wrong and changing this call breaks something, you could achieve pretty much the same result by just changing tokio's features from `"full"` to `"time"` in Cargo.toml.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1378,6 +1378,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "thiserror",
+ "tokio",
  "toml",
  "xdg",
 ]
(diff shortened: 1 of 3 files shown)


RobertMueller2:
I'm not going to merge this. Here's why:

I agree this won't make much of a notable difference if any, because the sleep amount is so small. And the question whether pulling in all these additional dependencies just for this small bit is certainly valid.

However, `std::thread::sleep(...)` blocks the thread it is running in, while `tokio::time::sleep(...).await` does not. There's a reason why async runtimes like tokio provide async sleep. So IMHO the question is not whether this breaks anything, but rather what is sensible in an async block.

I'm happy to think about reducing tokio's features from full to time for now. At first glance this reduces the amount of compiles by 9 crates, which isn't bad, and it still works. Chances are though, if we ever add additional async blocks, we might need full again.

RobertMueller2:
Thanks!

PR #230, opened 2025-06-01
fix: text: use font metrics instead of text metrics

### LINKED ISSUE #222: bug: text tool broken by multiple consecutive newlines
**To reproduce**:
1. Open satty
2. Select Text tool
3. Enter some text
4. Hit `Shift-Enter` multiple (2+) times
5. Only one newline appears, attempts to move cursor past the newlines lead to weird issues with cursor placement

**Expected behavior**:
Multiple consecutive newlines are entered into the text field, cursor behaves as usual

**Satty version**:
0.19.0 45741b69a8efc60bb735ec256df3c9d90d8edda6

### LINKED ISSUE #225: Text tool: Newline height inconsistent
I have no idea how I triggered this. But 1 and 2 are both showing space between baseline and next line lower case character, I believe they should be identical.

Image

RobertMueller2:
Closes #222
Closes #225

Text metrics does not respect stand alone newlines (values close to 0), might get rounded down (speculation).

Also, editing text (e.g. adding j or g on first line) move consecutive lines, gaps between lines can differ based on chars entered.

`font_metrics.height()` is identical with ascender - descender. This respects manual multiple newlines as well as offer consistent results for different glyphs -- unless glyphs are entered that use a fallback font. The latter is tricky to cover, GTK has the same challenge (see
https://gitlab.gnome.org/GNOME/gtk/-/issues/7555#note_2445986).

### DIFF
--- a/src/tools/text.rs
+++ b/src/tools/text.rs
@@ -65,13 +65,13 @@ impl Drawable for Text {
 
         let lines = canvas.break_text_vec(width, text, &paint)?;
 
+        let font_metrics = canvas.measure_font(&paint)?;
         for line_range in lines {
             if let Ok(text_metrics) = canvas.fill_text(self.pos.x, y, &text[line_range], &paint) {
-                y += text_metrics.height() * 1.2;
+                y += font_metrics.height() / canva_scale;
                 metrics.push(text_metrics);
             }
         }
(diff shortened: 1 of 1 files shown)


fabienjuif (approved):
Nice finding!

Issue #42, opened 2024-02-07
New Feature: Show Toolbars Only When The Mouse Hovered In


mistgc:
The toolbars don't display if the mouse doesn't hover in the app window.
<img src="https://github.com/gabm/Satty/assets/61856080/b9d6f49b-abf4-4307-abb6-b88014c056e6" width="300" height="200"/>
And show the toolbars when we hover in.
<img src="https://github.com/gabm/Satty/assets/61856080/15ddaded-0df0-4980-a904-c785b6ce4fe7" width="300" height="200"/>

gabm:
thank you for that idea.. could you explain what we try to achieve with this feature?

mistgc:
Imagine a scenario where a programmer is writing code while looking at other people's code snippets.
Like this:
![image](https://github.com/gabm/Satty/assets/61856080/7be3f16e-be90-4afc-863b-57d7997d1136)

mistgc:
We could pin some screenshots on the screen.

mistgc:
While this is true, we can add annotations to the image.

gabm:
okay.. well I think there is nothing wrong with hiding the toolbars on demand. I would prefer a (non-default) shortcut for this, like `Ctrl-T` for "toggle toolbar" or something like this.

I must say that I am a big fan of explicit behaviour (shortcut/button toggles toolbar) and not dynamic behaviour (mouse hover toggles toolbar). Further I think that should be "opt-in" .. not the default behaviour. There might be a setting "hide-toolbar-on-startup" or something similar..

I am happy to accept PRs

mistgc:
I implemented the basic functionality, but there isn't support for configuration.

mistgc:
Now, it supports configuration.
BUT, it doesn't support hovering to show toolbars.

KAGEYAM4:
What are your thoughts on Krita like toolbar which gets opened using right-mouse-button ?
![image](https://github.com/gabm/Satty/assets/75798544/9c6f7a3a-1fb3-4ae0-81eb-2096354de010)

mistgc:
I think most people find this very convenient, but for users who frequently use keyboard shortcuts (such as Window Manager, Vim-like Editor, Emacs), it is not convenient.

KAGEYAM4:
You're right, Plus it's will increase development time which would be wasteful as this dosen't offer much as keyboard based implementation will still be there plus the hovering toolbar.

alteriks:
[...]

PR #238, opened 2025-06-08
feat: show/hide toolbars based on window focus

RobertMueller2:
Addition for #42.

This does not affect the initial state of toolbars, only subsequent enter/leave events.

Further discussion might be needed in #42 whether default-hide-toolbars should have a tri-state (along the lines of "initial state of toolbars"), hide/show/focus-based.


fabienjuif:
That's cool!

PR #239, opened 2025-06-08
feat: ~/ replacement with home dir using dirs

### LINKED ISSUE #63: Support for ~/ in output dir
It seems that when using ~/ in the output dir path, no screenshot is saved and no error message is produced.

It would be good to be able to use the ~ shortcut for the home directory of the user.

RobertMueller2:
Closes #63

Simple implementation, decided against shellexpand because it seems unmaintained, whereas dirs moved its repository but seems maintained.

Suggest followup issue for further replacements discussions, this probably requires to move everything out of `handle_save()`.

Somebody should test this on Windows, if Satty ever runs there ;)


gabm:
we already have a dependency for xdg dirs.. could we only use one?

RobertMueller2:
Well, I don't feel like changing the other places.

`std::env::home_dir()` should be a reasonable replacement, though, it even has fallback. Perhaps it's best to use just that.

fabienjuif:
Create an issue maybe @RobertMueller2 (to migrate to XDG?)
Until then, I think this PR is creating value so I am :white_check_mark: and let you decide if we merge (I vote for a merge if it was not clear)

RobertMueller2:
> Create an issue maybe @RobertMueller2 (to migrate to XDG?) Until then, I think this PR is creating value so I am ✅ and let you decide if we merge (I vote for a merge if it was not clear)

For the home dir, I don't think using xdg adds any value over std. So I'd probably leave it at that in this code place. Std should always be ok, IMO.

I think @gabm was against using both xdg AND dirs as the latter also provides xdg paths, which is a valid concern. And for that, replacing the existing code bits using xdg with dirs, well, felt a bit pointless :)

gabm:
- yes, I meant what @RobertMueller2 said, he removed the new dependency and relied on std-lib stuff.. great!
- replacing xdg-directories might not be necessary, it just works ™️; therefore a new issue is not needed
- fyi: `std::env::home_dir` had been deprecated until rust 1.85 where it was "undeprecated" .. so its a future-proof way forward; via https://github.com/rust-lang/libs-team/issues/372

PR #246, opened 2025-06-18
feat: output-use-filepicker option to use instead of output-filename

TechHog8984:


RobertMueller2:
Thanks for the contribution. To be honest, I got mixed feelings about this.

Firstly, a file dialog is a great addition. We must have some issue somewhere that asked for it as well. ;) And rfd certainly has merit, and it's not a huge change within Satty. But there are also two disadvantages:
- additional dependency, when we use GTK4 which already has a cross platform file dialog. 
- the dialog shows up with an app_id of xdg-desktop-portal. AFAICT, rfd doesn't provide a way to set the app_id. And reparenting doesn't look straight forward, because you'd need to acquire a window handle from within handle_save. In my setup, Satty is floating and the file dialog is tiled, and no easy way to address this because xdg-desktop-portal is too generic as an identifier. This is inevitably going to trigger bug reports.

I'm also not too happy about `output-use-filepicker` cancelling out `output-filename` entirely. How Satty behaves or should behave if both options are present is very much a matter of interpretation, and again, this is going to trigger questions and bug reports. I'd prefer 
- either having `output-use-filepicker` use the `output-filename` as a default choice if both are present
- or having no additional config option for the dialog at all, and instead just show the file dialog whenever trying to save and `output-filename` is empty.

Personally I prefer the latter, but maybe we can listen to some additional input on this. I think on the last point we also need to ask @fabienjuif for his input, maybe he sees potential for an action?

gabm:
thanky for your contribution, Its a great addition.

Just as @RobertMueller2 said, I think we should use GTK4/Relm4 capabilities instead of another dependency. It also makes sure that it fits nicely with the existing mechanisms. https://relm4.org/docs/next/gtk4/struct.FileDialog.html

I think we will eventually have file dialogs for both input and output at some point, which are permanently available (open for discussion @RobertMueller2 @fabienjuif )

A button/shortcut should be available to trigger the file dialog for open (replaces the current image) and save at all times (defaults to the output-filename, if any), like so

- if an input filename was specified, Satty starts as usually, the filepicker is still available for opening and replacing another image
- if no input filename was specified, Satty starts and presents a file-picker before the main UI
- if an output filename was specified, Satty saves the image as usually. Additionally, a "Save As.." functionality is available through the menu/button/shortcut and Satty will use the file-picker for that. 
- if no output filename was specified, Satty always presents the file-picker, even on the current save button

fabienjuif:
@RobertMueller2 
> maybe he sees potential for an action?

This is not the same semantic IMO.

@gabm 
> [...]

I understand what  you try to achieve, while the balance between a not cloaked, lightweight tool and a feature rich tool is hard to find.
For example with what you are proposing, it will creates 2 more buttons, we are already discussing of having too much buttons, etc.

I feel like we want satty v2 that is widely different from what we have right now.
Maybe it's worth having a roadmap for this satty v2, while keeping a KISS philosophy for the current version

gabm:
this is a discussion that goes deeper than this PR though, right? 

What we experience is feature creep without a very clear vision of what we want and what we don't want.. this needs to be solved , even for a v2 .. i am open for discussions, but I'd like to have those in a structured way

fabienjuif:
For this PR I think this is reasonable to do what's @RobertMueller2 is mentioning:
> or having no additional config option for the dialog at all, and instead just show the file dialog whenever trying to save and output-filename is empty.

This is the least intrusive modification, while answering the feature don't you think?

gabm:
yes, i think so

TechHog8984:
Thank you all for the feedback. When I made the PR, I had really just wanted to add this feature in quick 'n dirty since a member of a Linux community discord server and I were discussing screenshot solutions and he liked my setup involving Satty but brought up the lack of a file dialog. I had recently used rfd for another project of mine so I figured it was an easy addition, but in hindsight I should have, as Robert mentioned, spent a little more time learning about the project before making a PR.

Anyway, I like Robert's solution discussed above and will work on implementing that (if I can figure out how to use git to update the PR). I also like the idea of using output-name as the "placeholder" so I might investigate that as well.

Thanks again.

TechHog8984:
I have given this an attempt, but I am not proficient enough in Rust to figure out how to use the gtk::FileDialog result, so I am going to personally drop development of this feature. I would be delighted to see someone pick this up.

w-jablonski:
I'd like to share a view. 

Perhaps an `Open File` filepicker is unnecessary and therefore could be cluttery. Because the user can always open a file with Satty from their file manager. It's exactly the same. Personally I never open an _empty_ image viewer to then pick an image via filepicker (but maybe others do...). Lastly, opening files from disk is not exactly screenshot app's main workflow anyway.

However, `Save As` is kinda necessary when editing files from disk. Coz when editing `Me_with_Monica.png`, or even annotating the yesterday's screenshot, you don't really want to save it as `%Y-%m-%d_%H%M%S.png` but usually to replace the original file. An alternative here would be to `Save Changes` but `Save As` is more versatile. Plus, it's _kinda_ still in line with screenshot app's workflow.

I admit that I present my views in stronger words than any of you guys, despite having contributed nothing. Please don't take offence, I'm just sharing my opinion.

wind-mask:
I think providing a `save_as` button first would be a quicker and more convenient goal.

RobertMueller2:
> I have given this an attempt, but I am not proficient enough in Rust to figure out how to use the gtk::FileDialog result, so I am going to personally drop development of this feature. I would be delighted to see someone pick this up.

Thank you for trying.. This PR has inspired #292, which achieved the same goal. I'm closing this PR now.

PR #252, opened 2025-07-30
option for filling shapes by default

oakrotka:
This PR add the `default-fill-shapes` option, which makes Satty fill the shapes by default. I also updated the readme to be up to date with the current help output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ save-after-copy = false
 default-hide-toolbars = false
 # Experimental: whether window focus shows/hides toolbars. This does not affect initial state of toolbars, see default-hide-toolbars.
 focus-toggles-toolbars = false
+# Fill shapes by default
+default-fill-shapes = false
 # The primary highlighter to use, the other is accessible by holding CTRL at the start of a highlight [possible values: block, freehand]
 primary-highlighter = "block"
 # Disable notifications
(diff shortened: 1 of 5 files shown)


RobertMueller2:
Thank you for the contribution. Other than the readme remarks above, it looks good to me. The lint errors have nothing to do with this PR, we'll address them separately. I'll open an issue for it.

Generally spoken, I know it's tempting to fix all of the readme while editing it anyway. But personally I'd prefer a separate PR for this. Reason being, if we have to revert a PR for whatever reason (not very likely for this one, but generally it's a possibility) we'd lose the readme fixes.  You don't need to change this now, just a remark for the future. We have to write this down as contribution guidelines at some point.

RobertMueller2:
regarding the readme comments, I think we have copied the help output to the README in the past. Perhaps we need to do that again (and perhaps automate it), and then add the experimental flags with the parameters or the help text.

oakrotka:
Should I revert the changes to the readme (except for the new option), then? I basically just copy-pasted the current help output there, because it changed with the new option and I wanted the readme to be accurate to the new output; fixing stuff broken in previous commits was just a side effect of that

RobertMueller2:
> Should I revert the changes to the readme (except for the new option), then? I basically just copy-pasted the current help output there, because it changed with the new option and I wanted the readme to be accurate to the new output; fixing stuff broken in previous commits was just a side effect of that

Yeah, I think that would be easiest, thank you.

We'll have to fix the readme in a separate issue.

Issue #116, opened 2024-08-25
[bug] Inserting a new line with Shift+Enter results in "no glyph" character


fnune:
See:

![Screenshot from 2024-08-26 01-48-10 (edited)](https://github.com/user-attachments/assets/fe244ef8-9d08-4fda-bcb3-eeb63a228a57)

fnune:
Update: looks like this only happens because I'm using a non-default font.

```toml
[font]
family = "Inter"
style = "Regular"
```

I would expect this to work anyway.

RobertMueller2:
I can reproduce it with that font. This will probably require a bug report for femtovg.

RobertMueller2:
femtovg fixed this upstream with release 0.14.1, we just pulled in a newer version via #267.

I can't reproduce the issue any longer. This is not in a satty release yet but git main head. Can you confirm when you get a chance? Otherwise we'll close this after next release with a bit of a delay.

RobertMueller2:
fix included in 0.20.0.

PR #272, opened 2025-09-25
feat: single-key and number-key tool shortcuts

structwafel:
Based on the work of @gabrielgcma and their PR:  #148

### Things done on top of their:
- merged main
- added shortcut keybinding configuration

### Things changed of their PR:
- let gtk handle ui update

I'm not familiar with gtk / relm4, but llm helped make [commit](https://github.com/gabm/Satty/commit/e634d88ee28905dbe163fb0043b17f6de55c95db) to solve the issue where changing tool with shortcut made the `initial_tool` still seem selected.
[Kooha-2025-09-25-05-37-23.webm](https://github.com/user-attachments/assets/93f770b7-3d9c-4d06-8d78-bd98d08652b4)

## Testing
Have tested using keybindings, and modifying `config.toml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,20 @@ All configuration is done either at the config file in `XDG_CONFIG_DIR/.config/s
 - `Ctrl+Y`: Redo
 - `Ctrl+Z`: Undo
 
+#### Tool Selection Shortcuts (configurable)
+Default single-key shortcuts:
+- `P`: Pointer tool
+- `C`: Crop tool
+- `B`: Brush tool
+- `I`: Line tool
+- `Z`: Arrow tool
+- `R`: Rectangle tool
+- `E`: Ellipse tool
+- `T`: Text tool
+- `M`: Marker tool
+- `U`: Blur tool
+- `G`: Highlight tool
+
 ### Tool Modifiers and Keys
 
 - Arrow: Hold `Shift` to make arrow snap to 15° steps
@@ -103,6 +117,20 @@ right-click-copy = false
 no-window-decoration = true
 # experimental feature: adjust history size for brush input smooting (0: disabled, default: 0, try e.g. 5 or 10)
 brush-smooth-history-size = 10
+
+# Tool selection keyboard shortcuts
+[keybinds]
+pointer = "p"
+crop = "c"
+brush = "b"
+line = "i"
+arrow = "z"
+rectangle = "r"
+ellipse = "e"
+text = "t"
+marker = "m"
+blur = "u"
+highlight = "g"
 
 # Font to use for text annotations
 [font]
--- a/config.toml
+++ b/config.toml
@@ -24,6 +24,20 @@ primary-highlighter = "block"
 # Disable notifications
 disable-notifications = false
 
+# Tool selection keyboard shortcuts (optional, defaults shown)
+#[keybinds]
+#pointer = "p"
+#crop = "c"
+#brush = "b"
+#line = "i"     
+#arrow = "z"    
+#rectangle = "r"
+#ellipse = "e"
+#text = "t"
+#marker = "m"
+#blur = "u"
+#highlight = "g" 
+
 # Font to use for text annotations
 [font]
 family = "Roboto"
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashMap,
     fs,
     io::{self, Write},
     path::Path,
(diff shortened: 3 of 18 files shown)


RobertMueller2:
Thank you very much for picking this up. Functionality looks kinda good at first glance. Few remarks from my side:

* the config allows me to state e.g. `arrow = "abc"`, satty doesn't complain about it, but the binding is unusable. I see some value in using a string in `KeybindsFile`, perhaps we might want to allow modifiers down the road one way or another, so the `Option<String>` can remain. But the shortcut itself should be a `char` or `gdk4::keys::Key` (or in the future maybe a struct combining `gdk4::keys:Key` and modifiers), and if it can't be parsed from the config file, there should IMO be a warning on stderr, and satty should use the default as a fallback.
* I would have preferred rebasing the initial commit onto main head rather than a merge commit. With a rebase, it's easier to see if any code is being changed that's outside of the PRs scope. But please don't bother changing that now, as I have already confirmed the merge commit is OK, that would be unnecessary work ;)
* if I use the text tool and press esc once, I can no longer change the active tool via shortcut
* if I use the s/m/l buttons, I can no longer change the active tool via shortcut
* ~there's one comment I need changed please, see inline review~

structwafel:
@RobertMueller2 
* char is much better + added error printing. I looked into making it a `gdk4::keys::Key`, technically could be done. But the [handle_text_commit](https://github.com/gabm/Satty/pull/272/files#diff-2c950573b5529d539f261ae6a60f0bf07d7346a58187b9fb8f1b02504a22dffeR484) function uses this `TextEventMsg::Commit(txt)` which is a string, but from the comments of the original pr it only allows single characters?
I'm not well versed with the `gdk4`/`relm4` system, but I think for future work the Key+modifiers could be used if some more work upstream the calling is done. But the input handling itself might not be ideal? As the text input doesn't handle spaces nice either? (only showing them after pressing other letter)

* That makes complete sense, wrong of me. Will organize better next time sorry! 😬

* should be fixed with: [`f6ede5c` (#272)](https://github.com/gabm/Satty/pull/272/commits/f6ede5cbbf5655b00a0fe5974c0ebe5b13b82e7f)

* I don't have this issue? I did all permutations I could think of?
satty > s/m/l button  > shortcut  == works
satty > shortcut brush > using brush > s/m/l > shortcut elipse  ==  works
satty > text > typing > s/m/l > typing > enter > shortcut  ==  works

What operations did you do for it to not work?

RobertMueller2:
> * I don't have this issue? I did all permutations I could think of?
> satty > s/m/l button  > shortcut  == works
> satty > shortcut brush > using brush > s/m/l > shortcut elipse  ==  works
> satty > text > typing > s/m/l > typing > enter > shortcut  ==  works
> 
> What operations did you do for it to not work?

Yeah, good question actually. 😄 I couldn't reproduce when trying again, so I guess we can disregard this one.

RobertMueller2:
> @RobertMueller2
> 
> * char is much better + added error printing. I looked into making it a `gdk4::keys::Key`, technically could be done. But the [handle_text_commit](https://github.com/gabm/Satty/pull/272/files#diff-2c950573b5529d539f261ae6a60f0bf07d7346a58187b9fb8f1b02504a22dffeR484) function uses this `TextEventMsg::Commit(txt)` which is a string, but from the comments of the original pr it only allows single characters?

I've had a look and didn't have a great understanding of this part yet. It seems to work when testing, but I'll try to get a better clue about it when I get a chance. I don't think we have any problem here, just want to make sure. Sorry for the delay ;)

>   I'm not well versed with the `gdk4`/`relm4` system, but I think for future work the Key+modifiers could be used if some more work upstream the calling is done. But the input handling itself might not be ideal? As the text input doesn't handle spaces nice either? (only showing them after pressing other letter)

I have tried to improve text input behaviour on the back of #116/#223. I believe this was covered by it. We ran into a femtovg issue which was reported, I felt there was no point trying to fix it before femtovg. Femtovg fixed it in the meantime, I think we updated the dependency too, so I should have a look at this again time permitting.

structwafel:
> I've had a look and didn't have a great understanding of this part yet. It seems to work when testing, but I'll try to get a better clue about it when I get a chance. I don't think we have any problem here, just want to make sure. Sorry for the delay ;)

No worries, I also didn't really understand either. Was happy that someone started this feature hahaha.

RobertMueller2:
That section is partially related to IMContext which was removed further up. The rest works, we know that already, but we should make sure the IMContext also still works, even though it may be replaced by #273 soon. I need to find out how to test this.

I think then we should be good to go.

RobertMueller2:
Have to cherrypick f6e183e, sorry. I'm stuck with wsl for a few days and can't get fcitx to work in there, so won't be able to check the IM functionality for a while longer.

RobertMueller2:
I can use fcitx input method with both before this change and after. I can not get it to work without explicitly adding a chinese (or other) font within text.rs, but there may be a font configuration that works from config. I'm not sure, I honestly don't know enough about this area.

What I can say though is that I can't find any indication this PR breaks it, or substantially changes anything regarding IM input.

Hopefully we can get #273 to improve the situation for everyone.

I'll merge this now, lgtm.

gabm:
wow, this has been a longstanding feature wish by everyone! Thanks so much @structwafel and @RobertMueller2 for getting this over the line 🥳

Issue #223, opened 2025-05-25
bug: text tool: cursor issues with multi-line text


postsolar:
**To reproduce**:
1. Open satty
2. Select the text tool and input this
```
aa
aa
aa
```
3. Press `Left` twice. The cursor is supposed to be on line 3 char 1 (3:1) but it's at 4:1.
4. Press `Left` once more. The cursor is supposed to be 2:3 but it's still at 4:1. After yet another `Left` it takes the correct position of 2:2.
5. Press `Left` once more. The cursor is again at 4:1. From thereon it's the same as in previous step.

**Expected behavior**:
Cursor behaves the same as in web browser input fields.

**Satty version**:
0.19.0 45741b69a8efc60bb735ec256df3c9d90d8edda6

RobertMueller2:
Yes, confirmed.

I've started to try and get an understanding what's needed to fix this. Then I reported another bug to femtovg on the back of #116, and I think this may change the way femtovg deals with newlines. Not sure if this affects the code we got, but it might. I don't think I'll work on this until we know more.

PR #273, opened 2025-09-26
Add IME preedit support, caret location updates, and CJK font fallbacks

agata:
## Summary
- Integrates GTK IME preedit lifecycle (start/changed/end) for the Text tool.
- Reports the live caret rectangle to GTK so IME candidate windows appear at the correct position.
- Improves font loading via fontconfig and adds a fallback stack for CJK scripts to reduce missing-glyph issues.

image

## Key Changes
- IME integration and caret positioning
  - src/main.rs: wires a shared `gtk::IMMulticontext`, forwards preedit events, and updates caret location via `set_cursor_location`.
  - src/sketch_board.rs: adds `SketchBoardInput::ImeCursorRect` and `SketchBoardOutput::UpdateImeCursor(Rectangle)`, plus helpers to convert image-space caret to widget-space `Rectangle`.
- Text tool preedit handling
  - src/tools/text.rs: implements preedit lifecycle:
    - PreeditStart/Changed/End with temporary insertion, cursor movement within preedit, and proper commit.
    - Ensures preedit is cancelled or committed when leaving/confirming the tool (Enter, Escape, tool change).
    - Computes caret geometry during editing and exposes it for IME positioning.
- Font loading and CJK fallbacks
  - src/femtovg_area/imp.rs: uses fontconfig to load the configured family; falls back to bundled Roboto if unavailable; adds a list of CJK fallback families and prevents duplicates.
  - src/femtovg_area/mod.rs: introduces a global font stack and uses it in the Text tool so `Paint::set_font` prefers the configured font but falls back gracefully.
- Coordinate utilities and minor fixes
  - Adds image↔canvas/widget conversion helpers used for caret rectangle mapping.
  - De-duplicates IME cursor notifications by caching the last rectangle.
  - Small fix: clone sender before passing (avoid move).

## Behavior Changes
- When using an IME (e.g., Japanese, Chinese, Korean) in the Text tool:
  - Preedit composition is shown and updated inline; caret moves according to the IME’s cursor position.
  - Candidate popup follows the on-canvas caret.
  - Confirming text commits preedit; cancelling or switching tools ends preedit cleanly.
- Text rendering uses a multi-font stack to reduce tofu for CJK characters.

## Testing Notes
- Start the Text tool and type using an IME:
  - Preedit composition appears inline; candidate window tracks the caret.
  - Move the IME cursor within the composition; caret updates correctly.
  - Press Enter to commit, Escape to cancel; switching tools ends preedit.
- Enter CJK characters without the configured primary font installed; verify glyphs render via fallbacks.

## Implementation Highlights
- Files touched:
  - src/main.rs
  - src/sketch_board.rs
  - src/tools/text.rs
  - src/femtovg_area/imp.rs
  - src/femtovg_area/mod.rs

## Notes
- Font fallback list is a best-effort set of common CJK families; availability depends on the system.
- No config migration needed; uses existing `font.family` when present and falls back otherwise.
- Sometimes the candidate window appears at the top-left corner of the display instead of near the caret. I found it difficult to resolve this myself. However, making it possible to input text with an IME at all is already a big step forward for IME users.
- For IMEs other than Japanese or Chinese, there will likely be missing fallback fonts. The font specified in family = "Your font name" will always be used with the highest priority.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "cargo" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,54 @@
+CONTRIBUTING
+==
+
+Contributions are welcome. Satty is not able to evolve without relying on its contributors and their contributions.
+
+This documents tries to reduce friction when it comes to contributions by defining some guidelines, some of which may follow a rationale while others are arbitrary determinations.
+
+Please note that opening a PR or even just an issue may expose your work to pertinent discussion regarding code quality, Satty's scope and these guidelines, and possibly things we haven't yet thought of. This isn't meant as discouragement, just as a heads-up.
+
+Issue first, then PR
+--
+
+The issue should state what is missing from or broken in Satty. All the discussion around whether a feature is in scope, or a behaviour is a bug can take place there. A related PR is then just about correctness of a fix or feature implementation. This ensures that a specific feature or fix is actually wanted.
+
+3rd party crates
+--
+
+We would like to keep 3rd party dependencies to a minimum. Addition of new dependencies should only be considered if
+- the relevant code parts are non-trivial
+- the functionality in question cannot be provided via existing dependencies
+
+Code comments
+--
+
+Ideally, code should be written in a way that it is self-explanatory. Comments can always help make code parts more understandable. They especially make sense when a section
+- was tricky to figure out
+- is sophisticated or unintuitive or not immediately obvious
+- might be in jeooardy of being overwritten by future you or other contributors due to not understanding it properly
+
+Please note that we may ask for additional comments.
+
+Code formatting and hints/improvements
+--
+
+Please use `cargo fmt` to apply formatting and `cargo clippy` to fix all suggestions pertinent to your changes. You can use `make fix` for both. Please note that this may apply changes unrelated to your code:
+- formatting if previous commits have not used `cargo fmt`
+- hints if previous commits have not used `cargo clippy` OR clippy is newer than the last commit and has learned new hints in the meantime
+
+Missing formatting/hints that precede your PR should be addressed via a separate issue/PR in main branch first. If in doubt how to resolve such a situation, ask.
+
+README changes
+--
+
+If a PR changes Satty's behaviour and where appropriate, please adjust `README.md` as well. `make update-readme` adds the command line help (output of `satty --help`) automatically which is relevant whenever command line arguments change. While it can be tempting to add other fixes to the README while you're at it, unrelated changes to it which precede your PR should be addressed in a separate issue/PR first. If in doubt how to resolve such a situation, ask.
+
+GenAI usage
+--
+
+GenAI usage is tempting and can save time, but it's not without pitfalls. At this point in time, full vibe coding mode can and often does lead to bad quality code which we are not going to merge.
+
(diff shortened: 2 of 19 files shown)


RobertMueller2:
Thanks a lot for the contribution.

It would be good if someone could have a look at this who has a minimum understanding of the subject matter, I can't say I do at this point.

I do have a remark and a heads-up though.

Question:
If the fallback font list contains more than just CJK fonts, we should probably name the variable something more generic. And perhaps the list of fallback fonts is something that actually should be configurable from outside, it could default to the list you provided.

Heads-up:
Both #272 and a fix for #223 are going to overlap. #223 might take longer, perhaps it's even easier to do after these changes, but I'm looking to merge #272 as soon as I get a chance to finish review. There is a chance some sections of this PR will have to be rebased accordingly. I know that rebasing can suck, so want to be transparent here.

agata:
Thank you for the review!

I’ll update the variable name for the fallback font list as you suggested.

Regarding where to place the fallback list: my initial thought was to embed it in the code, since if we move it to the config, existing users would need to update their own config manually. That said, I agree with your point that loading from the config would be more flexible and keep the code cleaner. Another simple approach would be not to support a fallback list at all, and instead let users set the desired language directly via `family = "Roboto"` in `config.toml`.

So the options I see are:
A. Keep an internal fallback font list
B. Load a fallback font list from `config.toml`
C. Don’t support a fallback list and just let users set `[font] family` directly

Personally, I’m leaning toward option C and would like to test if this works reliably.

Thanks also for pointing out the potential conflicts with other PRs.

For testing, I’m currently using fcitx5 + mozc for Japanese. It would be great if others could confirm with different IME systems or languages.

agata:
I’ve confirmed the latest main branch after #272 was merged, and Japanese text input now works correctly. I’ll close this branch.

That said, I see a few areas in the current IME support / text input behavior on main that could be improved. I’ll experiment with these in a separate branch:

* Inline rendering of uncommitted IME text (to make the typing UX closer to common applications)
* Caret vertical size when typing with large font sizes
* Line spacing when pressing Enter with large font sizes

RobertMueller2:
> I’ve confirmed the latest main branch after #272 was merged, and Japanese text input now works correctly. I’ll close this branch.

Thanks for confirming, that's good to hear. Just out of curiosity, which font configuration do you use? My gut feeling is that we can still improve fallback font handling, but I could be mistaken.
 
> That said, I see a few areas in the current IME support / text input behavior on main that could be improved. I’ll experiment with these in a separate branch:
> 
>     * Inline rendering of uncommitted IME text (to make the typing UX closer to common applications)

Yes, that would be a nice addition, I think.
 
>     * Caret vertical size when typing with large font sizes
> 
>     * Line spacing when pressing Enter with large font sizes

100%, these two are an issue even without IME support in mind. There's a gap between gtk and femtovg, line/glyph height retrieval is dodgy, and we have to take into account manual vs automatic line breaks. I think these two could actually be a separate fix/branch from the feature above.

agata:
Thanks for the feedback.

Regarding fallback fonts, I didn’t implement support for them. At the moment, users need to set the font-family to a font that supports the characters they plan to input via IME. Otherwise, they’ll see tofu (blank squares). I think it would be best to document this clearly in the README.

If we research which fonts are supported across major distributions and prepare a fallback that covers most languages, it would make things more convenient since users wouldn’t need to adjust font names manually. I might look into this when I have time.

As for the line spacing issue with large font sizes, I already addressed it in https://github.com/gabm/Satty/pull/283, though I made the fix together with the IME improvements (commits are separated). Do you think it would be better to split this into a separate pull request?

agata:
I am using the following font. It is a Japanese font.

```
[font]
family = "Noto Sans CJK JP"
```

RobertMueller2:
> As for the line spacing issue with large font sizes, I already addressed it in #283, though I made the fix together with the IME improvements (commits are separated). Do you think it would be better to split this into a separate pull request?

Generally spoken, smaller PRs can be useful if there is a chance they can be merged separately. E.g. in this case something is fixed that's not necessarily depending on the IME change. If e.g. we needed longer time to discuss details of the IME change, this would more or less block the other fix along with it. That would be unfortunate.

But in this particular case, we're almost at the finishing line with #283. So please leave it as it is. Splitting it now would be unnecessary work with no benefit.

PR #274, opened 2025-09-28
Improve ime support fallback

agata:


RobertMueller2:
Just a small hint ;) You don't really need a separate PRs to add more changes. When you update a branch on github which is already tracked by a PR, the PR automatically gets updated, too.

agata:
Thanks! You’re right—I understand now. I’m working on a revised version and will push an update soon.

PR #283, opened 2025-09-30
Improve IME Inline Rendering and Text Metrics

agata:
## Summary
- Inline rendering for IME preedit text on the canvas improves in-progress input feedback.
- Caret and newline heights now align with the active font metrics for consistent visuals.

image

image

## Key Changes
- `src/sketch_board.rs`, `src/tools/text.rs`: render IME preedit strings inline with proper highlight and positioning.
- `src/tools/text.rs`: recalculate caret height ahead of input using font-size-derived metrics.
- `src/tools/text.rs`: adjust newline insertion logic to match real font line height.

## Behavior Changes
- IME preedit text appears inline during composition.
- Caret height is correct before typing, and newline spacing follows the font’s true metrics.

## Testing Notes
- Manual: Verified Japanese IME preedit, commit, and newline insertion; confirmed caret and line spacing track the font size.

## Implementation Highlights
- Added helpers to draw preedit text in sync with existing text rendering (`src/sketch_board.rs`, `src/tools/text.rs`).
- Replaced hardcoded caret/newline heights with values derived from shared font metrics utilities (`src/tools/text.rs`).

## Notes
- Reuses portions of the previously closed PR #273.

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -67,6 +67,8 @@ pub struct KeyEventMsg {
 #[derive(Debug, Clone)]
 pub enum TextEventMsg {
     Commit(String),
+    Preedit { text: String, cursor: Option<u32> },
+    PreeditEnd,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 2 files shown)


RobertMueller2:
Nice! I won't pretend I have any idea how people normally use this, but personally I'm impressed.

Would you like to add a remark to the README? I could imagine
-  mention of default hotkeys "Ctrl+Space" to cycle input method, "Ctrl+Alt+P" to toggle preedit, OR a link to somewhere else that covers it.
- the need to specify a font that supports it (for now)

I'm sure people familiar with IME know all this already, just for completeness sake. 

I've also noticed a few things:
- When using e.g. Noto Sans CJK and latin input, Satty can become quite sluggish. I suppose we have to accept this as a known issue until we get a better way to set fallback fonts down the road (could be related to #244)
- when using manual line wraps (shift+enter), cursor back/forth seems to get stuck for an additional char (the line break). This was already reported in #223, so it's not new and we can leave this open and fix it elsewhere, I think the caret behaviour is already an order of magnitude better than it was.
- auto wrapping can sometimes be beyond image bounds depending on window size (see screenshot). I could only reproduce with Latin input, I think that's new and should ideally be resolved before we merge.

Thanks!

satty-20251004-082526

RobertMueller2:
>     * auto wrapping can sometimes be beyond image bounds depending on window size (see screenshot). I could only reproduce with Latin input, I think that's new and should ideally be resolved before we merge.

Nope, not new. It's rendered properly when saved ... just it isn't exactly WYSIWYG. I'm assuming the width is the window rather than image width used with `canvas.break_text_vec`. I'll record a new issue for that.

lgtm.

RobertMueller2:
Oh, just forgot, can I please trouble you to create a followup PR for 1 or 2 lines in the README? Thanks :)

PR #284, opened 2025-09-30
Update Annotation Size Spinner Step

agata:
## Summary

Adjusted the annotation size dialog so the spinner buttons jump in whole-number increments instead of 0.01 steps. 

Most people tweaking annotation sizes are looking for a visibly larger or smaller mark, not a fractional change they can’t perceive. Switching the spinner to whole-number steps matches that expectation, gets them to the desired size faster, and keeps the dialog from feeling unnecessarily fussy.

## Testing Notes
* Manual: Opened the annotation size dialog and confirmed that the up/down arrows now change the value by ±1.

### DIFF
--- a/src/ui/toolbars.rs
+++ b/src/ui/toolbars.rs
@@ -746,8 +746,8 @@ impl Component for AnnotationSizeDialog {
 
                     set_tooltip: "Annotation Size Factor",
                     set_numeric: true,
-                    set_adjustment: &gtk::Adjustment::new(0.0, 0.0, 100.0, 0.01, 0.1, 0.0),
-                    set_climb_rate: 0.1,
+                    set_adjustment: &gtk::Adjustment::new(0.0, 0.0, 100.0, 1.0, 5.0, 0.0),
+                    set_climb_rate: 1.0,
                     set_digits: 2,
                     #[watch]
                     #[block_signal(value_changed)]


RobertMueller2:
OK, the defaults aren't good, you definitely got a point there. I'm not opposed to improving them. But I'm using decimal adjustments all the time, so I'm slightly reluctant to just overwrite it with an integer value. :smile: 

This definitely has some overlap with #178, #171, doesn't mean we need to wait for them, but there's a chance whatever we do here will be overridden in the future.

I'm not sure if we ever find a sensible default that works for all image sizes, so I'm kinda thinking, can we perhaps set the adjustment and acceleration depending on image size? (And yes, that will add an additional problem if we ever allow to open images after starting, but we can deal with that once we cross that bridge.)

agata:
I’ve checked #178 and #171, and I think they’re addressing the same underlying issue.
I like the idea of switching via a button — that’s a nice improvement. It would be even better if it could be changed without opening a dialog.

Some similar tools allow the entered text to become selected after input, so it can be moved or resized with a handler. Personally, I prefer that approach, though I imagine it would be more challenging to implement.

robertwidfen:
Please check out the merged PR #549. It removes the dialog and adds increments of 1.0, 0.1, and 0.01 steps (details are in the README.md). A middle-click on +/- uses a 1.0 step size, while the default is 0.1.

I noticed you added `set_climb_rate: 1.0` in your changes. I missed that in my PR, so I'll bring it back! However, 1.0 feels a bit too fast for me - I keep overshooting the target value. 😉

Also, I'm not entirely sure if a step size of 5.0 is intuitive here. What do you think?

PR #289, opened 2025-10-05
fix text tool cursor issues

RobertMueller2:
Closes: #223

After looking at this with a debugger, I do not agree that femtovg is not reporting line breaks in the byte index. The problem is rather that we cannot distinguish first char (index 0 + last line last index) in a new line from the last char in the last line (last line last index).

Debugger:
satty-20251005-072901

Behaviour after fix:
https://github.com/user-attachments/assets/e9db27f1-a189-445c-9be0-8770820c4767

Cursor height, positioning relative to char can still be tweaked on the back of a followup issue.


RobertMueller2:
Hmm... I think now the Home and End keys need some change though. Not that they worked great before, but let me try to improve that too ;)

Issue #92, opened 2024-06-14
[feature] Add option to choose location of where to save the file


sga-13:
The title. Can an option (either graphical or keyboard shortcut like Ctrl+Shift+S or something alike) which will open a file picker (maybe the gtk one or the xdg file picker) and we are able to choose a location to save the output, which may or may not be independent of where the Ctrl+S would save

I tried to achieve something similar by using yazi (a terminal file manager) to act as a file picker, and using that either choose the location to save or the exact file path, but somehow i am not able to stop it from rendering before the screenshot is taken

I have tried to 

```
grim -g "$(slurp)" - | satty --filename - --annotation-size-factor 0.4 --output-filename "$(yazi-filepicker)/out.png"
```

and other variations where i give full path, or add a sleep time (which works but limits me on how fast should i snip)

any help is welcomed, and if this issue is duplicate, then I am sorry, but I could not find any

gabm:
This feature is not present (yet). Satty started life as a CLI tool without graphical interaction apart from the annotation task itself. But it developed since and the lack of a graphical file picker for saving is something that is lacking - i fully agree.

To preserve the current functionality, I suggest the following mechanism:

- if the user specified a save path or pattern, then use that one as "is"
- if the user did not specify an output filename, then we show the button anyway (contrary to now) and upon button press, show the gtk file picker and then save the file there...

I think that would be quite easy to implement - do you want to take a shot? I can provide guidance

sga-13:
I have no experience in rust, all i can do is python and bash scripting (and weirdly java), I can try to take a shot at it, but please do not expect much

i am going to imagine that we include some function here (like color palette has something set by default, but instead calling in filepicker)

```
impl Default for Configuration {
    fn default() -> Self {
        Self {
            input_filename: String::new(),
            output_filename: None,
            fullscreen: false,
            early_exit: false,
            initial_tool: Tools::Pointer,
            copy_command: None,
            annotation_size_factor: 1.0,
            save_after_copy: false,
            color_palette: ColorPalette::default(),
            default_hide_toolbars: false,
            font: FontConfiguration::default(),
        }
    }
}
```

if this is correct, then I can try to take a shot, if not please guide me

nomm-dot:
[...]

PR #292, opened 2025-10-06
feat: add 'save as' button

Xismy:
Related to #92 and #246


RobertMueller2:
Thanks for the contribution.

Tested it and it looks good, there's just one minor thing, the "save as" button uses the same icon as "save". I'm not sure there's a lot of great alternatives in https://github.com/Relm4/icons/blob/main/icons16.md. Perhaps we could use "save-filled" or "save-multiple-regular" to distinguish? It would have to be added to `build.rs` in order to work.

I think in the future we may also want to consider adding that to `actions-on-...`, but we'll leave that discussion for a followup issue/PR.

RobertMueller2:
lgtm. Thanks!

PR #295, opened 2025-10-08
updated config.toml to be up to date with README.md

PascalKont:
I noticed the config.toml file has old content that throws an error when used.

Error reading config file: Decoding toml failed: TOML parse error at line 48, column 1
   |
48 | first = "#00ffff"
   | ^^^^^
unknown field `first`, expected `palette` or `custom`

So i updated the config to be in sync with the README.md

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -3,6 +3,8 @@
 fullscreen = true
 # Exit directly after copy/save action
 early-exit = true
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Thanks for catching that one. Hopefully with #200 we can automate this at some point ;)

Issue #291, opened 2025-10-05
Fallback font handling


RobertMueller2:
Right now, entering characters not in the specified font result in tofu or nothing.

Femtovg itself is capable of using fallback fonts, as shown in at least two of femtovg's examples. So it should be possible to improve this in Satty.

This might or might not work better if fallback fonts can be user configured. If so, we also need a sensible default configuration. And perhaps consider shipping suitable fonts.

PR #298, opened 2025-10-09
feat: add fallback fonts

RobertMueller2:
Closes: #291

@agata I hope you don't mind, I've extracted the fallback font handling from your earlier PR #273 and added configuration file bits. This seemed very useful. I'm leaving this in draft mode until you confirm it's ok. If you want, you can open your own PR instead. ~In the meantime, i'll see if I can use cherry-picking and rebase to make your work more visible.~

This could receive a default config, for now I just updated the configuration section of the readme.

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -186,6 +186,10 @@ impl FemtoVGArea {
             self.canvas.borrow_mut().replace(c);
         }
 
+        if self.font.borrow().is_some() {
+            return;
+        }
+
         let mut canvas_ref = self.canvas.borrow_mut();
         let canvas = canvas_ref.as_mut().unwrap();
 


RobertMueller2:
@agata
I've applied a bit of git magic to separate my changes from your changes. Please let me know if we can use your (slightly edited) commit like this.

agata:
@RobertMueller2 
Absolutely, that's perfectly fine and actually very helpful. Thank you for doing that. I appreciate it.

gabm:
where do these fonts come from? Are they supposed to be installed in the system?

RobertMueller2:
> where do these fonts come from? Are they supposed to be installed in the system?

Added clarification to README.

Can't comment on the source, but the fonts are not shipped and need to be available on the system. This PR doesn't provide a default setting, so the config in the README is just an example which can be refined at a later point in time. Or it may be superseded by #309 implementation before that, who knows. ;)

RobertMueller2:
converting back to draft, I want to have a closer look at how often we're loading fonts.

RobertMueller2:
Pretty much everytime `ensure_canvas` is used, which is frequent. This feels unnecessary, either we have something sensible at startup or we don't, but trying it again while Satty is running feels pointless at first glance. I'll try to improve this.

RobertMueller2:
I could have sworn I posted the two recordings here... probably some random issue on github now has a weird out of context comment by me but I have no idea how to locate it :p

Anyway, here's the recordings again. This is using #343, and this PR on top of it. I think it's quite a difference. Before:

https://github.com/user-attachments/assets/0bff6827-3f5b-4e24-b5be-e1df11561ab7

After:

https://github.com/user-attachments/assets/c06c5e67-b5c1-467d-bb48-632c4caa48f1

Issue #71, opened 2024-03-28
Feature request: Command-line parameter to override size (or perhaps disable resizing-on-init completely)


somebody1234:

gabm:
could you elaborate what behaviour you want to improve? how should it be?

somebody1234:
I want to be able to open a satty window that spans all monitors at the same time, but the hard limit of 80% of monitor size is preventing that.

For now I've just patched in a `--no-resize` parameter locally that immediately returns from the function that resizes the window, and I've instead set up my WM so that it resizes the window to the appropriate size

gabm:
Currently, Satty resizes to 80% because that makes the window floating by default on most WMs. That is the approach being used by swappy and has been adopted. Atm I see this as some kind of hack because I'm not sure this is even a "standard".

I see the possible uses for:

- no resize - let the wm decide what to do
- full screen on one monitor - this is what happens currently with `--fullscreen`
- full screen on all monitors - this is what you ask for. I am not sure there is a proper option in current UI toolkits. Afaik, most software focuses on single  monitors
- have it floating - this is the current default achieved by 80% resizing 


Any of these could also be achieved by a mixture of "no resize" and a program specific config in the config of your wm. 

that makes me think that no-resize might even be default, the floating or full screen just an option...

PR #302, opened 2025-10-10
feat: initial-size config and command line

RobertMueller2:
Closes: #71

Changes fullscreen (command line and config) to use a parameter:
- all (to cover all screens, probably not working in most compositors, but sets according property on toplevel. This is not the same thing as e.g. sway global fullscreen, btw)
- current-screen

Old fullscreen (`--fullscreen` and `fullscreen = true`) still working, this does the same as using "current-screen".

Adds option for initial resize
- smart (existing behaviour) (e.g. `--resize smart` or `resize = { mode = "smart" }`)
- actual size (e.g. `--resize 2000x800` or `resize = { mode = "size", width=2000, height= 800 }`)

Adds option for the existing floating hack (`--floating-hack` or `floating-hack=true`), default is off.

- [x]  WindowSizing in both command_line and configuration needs deduplicating
- [x] behaviour with different compositor settings for the window needs testing
- [x] reconsider default
- [x] consider all-screens
- [x] use options instead of None directly
- [x] consider changes from #335 once those are merged

~Requires #322 to be included before next release~.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ zoom-factor = 1.1
 text-move-length = 50.0 
 # experimental feature (NEXTRELEASE): Scale factor on the input image when it was taken (e.g. DPI scale on the monitor it was recorded from).
 # This may be more useful to set via the command line.
+# Note, this is ignored with explicit resize.
 input-scale = 2.0
 
 # Tool selection keyboard shortcuts (since 0.20.0)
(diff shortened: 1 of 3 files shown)


RobertMueller2:
I think I'll change this to
`--fullscreen=all|current|none` (with `--fullscreen` and `fullscreen=true` in config still working)
`--resize-window=w,h|smart`
`--floating-hack`

because these might be used in any combination. You could place Satty in Fullscreen and still want to have a floating, sized window if you exit fullscreen. I think this should work better for the control flow in `main.rs`.

EDIT: included above

RobertMueller2:
Changing back to draft so we can include #335 here.

RobertMueller2:
updated for PR #335.

Check errors may be temporary, will try again later.

Issue #303, opened 2025-10-11
command_line.rs belongs to build crate


RobertMueller2:
### Version

0.20.0 / 9792b1016c5a04d5280abb6553dc2ba6c4aa9181

### Description

Because `command_line.rs` is included in `build.rs`, it's actually placed in the build crate. (Not sure about the exact name of which).

This means, it is not in the satty crate. Which causes some funny effects such as not being able to use any crates from Satty's dependencies. Only after adding serde as a build dependency, I was able to import it in `command_line.rs`.

From what I've found via google the only way out seems to be placing `command_line.rs` in its own crate and use it both as a dependency and build dependency.

PR #304, opened 2025-10-12
fix: move command_line to separate crate

RobertMueller2:
Closes: #303

Currently, by including `command_line.rs` in `build.rs`, the former gets snatched into crate "build_script_build".

This has a number of funny and unintuitive side effects like dependencies not working for the file but the build dependencies.

Here we move `command_line.rs` into a separate "cli" crate. Technically, this might not be much of an improvement as we still have to add dependencies to the new crate, but it makes the behaviour more obvious.


RobertMueller2:
I'm kinda thinking the [clap documentation](https://docs.rs/clap_complete/latest/clap_complete/aot/fn.generate_to.html) should point this out, it does suggest `include!`. (Which caused the same result in my test) 

Or maybe it's a shortcoming of the rust compiler. My expectation would be that a source file could be part of both, of course in that case its dependencies would have to be in both dependencies AND build-dependencies. But perhaps my understanding of this isn't good enough at this point ;)

PR #308, opened 2025-10-14
Improve IME caret positioning and preedit rendering with Pango-enhanced text tool

agata:
## Summary
- Introduce a shared `IMMulticontext` so the text tool keeps caret focus in sync with IME sessions.
- Leverage Pango attribute parsing (via the new dependency) to render IME preedit foreground, background, and underline styling correctly.
- Rework line layout and caret math so wrapped lines, manual breaks, and IME cursor updates all land at the expected position.
- Set the context ID for fcitx/fcitx5 and forward caret rectangles to the IME to stabilize candidate popup placement.

https://github.com/user-attachments/assets/26b9f711-57e9-4011-a19d-6a70f146bebd

## Testing
- [x] cargo test
- [x] Manually enter text with an IME (e.g. fcitx5/Japanese) and confirm caret and preedit styling look correct.

### DIFF
--- a/src/ime/mod.rs
+++ b/src/ime/mod.rs
@@ -1,0 +1,2 @@
+pub mod pango_adapter;
+pub mod preedit;
--- a/src/ime/pango_adapter.rs
+++ b/src/ime/pango_adapter.rs
@@ -1,0 +1,142 @@
+use glib::translate::FromGlib;
+use pango::{AttrColor, AttrInt, AttrList, AttrType, Underline};
+
+use crate::style::Color;
+
+use super::preedit::{PreeditSpan, UnderlineKind};
+
+#[allow(clippy::cast_possible_truncation)]
+fn to_style_color(color: pango::Color, alpha: Option<u16>) -> Color {
+    let to_u8 = |value: u16| -> u8 { (value / 257) as u8 };
+    let alpha = alpha.unwrap_or(u16::MAX);
+    Color::new(
+        to_u8(color.red()),
+        to_u8(color.green()),
+        to_u8(color.blue()),
+        to_u8(alpha),
+    )
+}
+
+fn clamp_index(index: i32, len: usize) -> usize {
+    if index < 0 {
+        0
+    } else {
+        (index as usize).min(len)
+    }
+}
+
+fn underline_from_pango(value: i32) -> UnderlineKind {
+    match unsafe { Underline::from_glib(value) } {
+        Underline::None => UnderlineKind::None,
+        Underline::Single => UnderlineKind::Single,
+        Underline::Double | Underline::DoubleLine => UnderlineKind::Double,
+        Underline::Low => UnderlineKind::Low,
+        Underline::Error => UnderlineKind::Error,
+        _ => UnderlineKind::Single,
+    }
+}
+
+/// Convert a Pango attribute list into neutral preedit spans understood by our renderer.
+pub fn spans_from_pango_attrs(text: &str, attrs: Option<AttrList>) -> Vec<PreeditSpan> {
+    let mut spans = Vec::new();
+    let text_len = text.len();
+
+    let Some(attr_list) = attrs else {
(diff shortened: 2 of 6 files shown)


lilydjwg:
The input window is at the correct position, but it is still needed to specify a font for rendering. Maybe you can use pango to draw text onto a `cairo` surface and then draw that onto `femtovg::Canvas`?

RobertMueller2:
> The input window is at the correct position, but it is still needed to specify a font for rendering. Maybe you can use pango to draw text onto a `cairo` surface and then draw that onto `femtovg::Canvas`?

that's expected until we merge #298, and with that it'll still be required to configure fallback fonts. This might still not be very convenient, but should be an improvement over now.

Using Pango would mean placing plain image output onto the canvas. That's not to say we can't do it, but that's a considerable change from what we have now with using femtovg text rendering.

RobertMueller2:
Mabye it's possible to have Pango reveal all fonts it used for a certain text, then try to load those to femtovg. This would be a replacement or enhancement of the fallback mechanism in #298. Big maybe though, just a thought ;)

An alternative to all that could be what the femtovg text example is using, cosmic_text.

EDIT: But that would be for a future change.

RobertMueller2:
And just a heads up, it'll probably take us a while to review this.

lilydjwg:
> Mabye it's possible to have Pango reveal all fonts it used for a certain text, then try to load those to femtovg. This would be a replacement or enhancement of the fallback mechanism in #298. Big maybe though, just a thought ;)

Yes, it is possible: https://docs.gtk.org/Pango/method.Fontset.get_font.html.

RobertMueller2:
> > Mabye it's possible to have Pango reveal all fonts it used for a certain text, then try to load those to femtovg. This would be a replacement or enhancement of the fallback mechanism in #298. Big maybe though, just a thought ;)
> 
> Yes, it is possible: https://docs.gtk.org/Pango/method.Fontset.get_font.html.

I've added a followup issue for this so we can have a look in the future.

RobertMueller2:
Thanks for this PR, looks like quite a bit of work went into it. I've taken the time to get a better understanding of it and I do have some remarks. :)

Firstly, I have two code comments that I'll add in place in a minute. (EDIT: actually, just one code comment :wink:)

For some input, some part of the letters are inside the box while others aren't. It's not a biggie, but in case you want to fix this, feel free. I'm trying to attach a video that shows it.

Also just to point out, we may at some point add something that indicates text selection. I hope we'll be able to distinguish text selection and active IME. But I'm perfectly fine with leaving that as a future problem.

Lastly, I'm a bit unsure if I like this use of pango just for storing a text attribute representation. My gut feeling is that in this case I would prefer to define the struct(s) for this ourselves and not use the additional dependency. Unless I'm missing something here.

https://github.com/user-attachments/assets/a9f193e1-65db-4b0f-943d-05a52478f44d

agata:
Apologies for the wait, and thanks a lot for the specific suggestions.

The inline comment is now addressed in the latest push. I also adjusted the foreground re-draw scissor to add a tiny padding (±1 px with ceil/floor). Could you please check if this resolves the clipped glyphs? I couldn't reproduce the issue on my end, so if you still see any odd cases, it might be an environmental factor specific to my font specifications.

Regarding the Pango concern, I completely agree that we shouldn’t drag Pango types through the whole pipeline. The current revision keeps Pango confined to a small adapter (ime::pango_adapter) that only runs inside the GTK IME handler. Everything downstream now uses the neutral Preedit/PreeditSpan structs in ime::preedit, so femtovg really just reads our own data.

With these changes in place, could you take another look when you have a moment?

RobertMueller2:
@agata no need to apologise. Thank you for the changes.

The glyphs can still be clipped if they go to ascent or descent, such as j,g,h,b etc. I'm just using default font. But honestly I don't think it's a big problem, I'm happy to leave this open for now.

lgtm, thanks!

PR #319, opened 2025-10-28
Add copy filepath to clipboard action

Iann29:
Adds a new action that copies the saved file path to clipboard. Useful when you need to quickly reference where the screenshot was saved.

- New action: copy-filepath-to-clipboard
- Keybind: Ctrl+Alt+C
- Auto-saves if file wasn't saved yet
- Works with wl-copy and native GTK clipboard

Can be used in actions-on-enter, actions-on-escape, etc.

Tested on my system and works well. Let me know if you'd like any changes!

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ All configuration is done either at the config file in `XDG_CONFIG_DIR/.config/s
 - <kbd>Ctrl+C</kbd>: Save to clipboard (may be masked by active tool)
 - <kbd>Ctrl+S</kbd>: Save to specified output file
 - <kbd>Ctrl+Shift+S</kbd>: Save using file dialog <sup>0.20.0</sup>
+- <kbd>Ctrl+Alt+C</kbd>: Copy last saved filepath to clipboard <sup>NEXTRELEASE</sup>
 - <kbd>Ctrl+T</kbd>: Toggle toolbars
 - <kbd>Ctrl+Y</kbd>: Redo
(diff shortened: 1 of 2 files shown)


RobertMueller2:
Nice idea, thanks! Might be a bit before we get round to reviewing this.

I'm wondering if this needs to be a dedicated action, perhaps we should just always copy the filename when saving to file? Does anyone see any reason why we wouldn't want to?

40024:
If both are coupled, I think it would be best if there was a config to split them
I personally can see myself saving to file and not wanting to update my clipboard

RobertMueller2:
OK, maybe a separate action is the way to go. 

@Iann29 
I do have a few remarks:

1. I think the action should do just that: Copy the last used filename to clipboard, and if there is none, do nothing (or maybe log/notify there is none, I'd be fine with that). No auto-save. This should simplify this change a bit.
2. I think `copy_text_to_external_process` and `save_to_external_process` can be deduplicated. This requires a small change to `handle_copy_clipboard`, but that'd be fine. Can you try, please?
3. It would be great if you could add the new shortcut to README.md, along with `<sup>NEXTRELEASE</sup>` (The latter creates a dependency on #322)
4. Please use `make fix` to make the fmt task happy. Thanks :)

RobertMueller2:
Pushed commits: 
- rebase to main
- apply changes requested above
- fixed readme

RobertMueller2:
Looks good to me now. Thanks for the PR :)

PR #320, opened 2025-10-30
Remove broken `--help` highlighting in README.md

TobiasDummschat:
# What?

Removed the `shell` syntax highlighting for the `satty --help` codeblock.

# Why?

The `--help` is mostly text, leading to an apostrophe rendering the second half to be highlighted as a string. There are also some undesired syntax highlighting like red keywords `if` and `for`.

2025-10-30T13:34:31+01:00

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ custom = [
 
 ### Command Line
 
-```sh
+```
 » satty --help
 Modern Screenshot Annotation.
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Thanks for spotting this. Looks like a case of _Betriebsblindheit_ ;)

Issue #318, opened 2025-10-28
Numbered Marker: Number unreadable if color is white


40024:
### Version

0.20.0-1

### Description

Satty-ception

<img width="662" height="423" alt="Image" src="https://github.com/user-attachments/assets/52cb77d0-8367-40c7-be9b-7790e345436d" />

Never touched Rust before, but this seems like a simple fix, just flipping inner color to black if main color is (or is close enough?) to white

Maybe I'll read up and pr this myself if that's welcome 👀

gabm:
thanks for you bug report. Indeed, looks like a simple fix. Algorithms for colour/contrast analysis should be findable online.

The draw function for the marker is here: https://github.com/gabm/Satty/blob/main/src/tools/marker.rs#L28-L86

As you can see, white is hardcoded as the text colour.

RobertMueller2:
Sounds reasonable. Doing the flip as such shouldn't be a huge problem, the white colour is hardcoded here:

https://github.com/gabm/Satty/blob/04f67b297e8d52233e2f38427b3df3428daa996e/src/tools/marker.rs#L36

The bigger challenge may be to determine whether `self.style.color` is closer to white or closer to black. I think the best would be to look at other open source programs how they're doing it.

Intuitively, I'd say add the RGB values and use the average. And then there's other information around like [this post](https://learn.microsoft.com/en-us/answers/questions/1497116/how-to-find-if-a-given-color-is-dark-or-light), talking about something like "perceived brightness".

PR #321, opened 2025-10-31
Change inner numbered marker color based on luminance

40024:
Issue: https://github.com/gabm/Satty/issues/318

image


RobertMueller2:
Pushed commit:

- reuse of new var `marker_color`
- add Wikipedia link to Luma (which contains pertinent Rec. 709) in comment so we have a rough idea where this may have originated from ;)

Looks good to me, thanks for the contribution.

40024:
Thanks, happy to contribute!

PR #330, opened 2025-11-01
fix(crop): return Unmodified on Escape when crop is inactive

### LINKED ISSUE #329: Escape key not triggering actions-on-escape after deactivating crop tool
### Version

satty 0.20.0-1 from arch linux 'extra' repository

### Description

When the crop tool is active and Escape is pressed to deactivate the crop, subsequent Escape keypresses don't trigger the configured actions-on-escape behavior, e.g. if set to 'exit' in order to close the app, it won't do that.

Steps to Reproduce:
1. Launch Satty: grim -t ppm - | satty --filename - --fullscreen --initial-tool=crop --actions-on-escape=exit
2. Press Escape once (crop deactivates)
3. Press Escape again

Expected: Satty exits
Actual: First Esc deactivates the 'crop' active mode, 2nd and subsequent Esc do nothing.

Workaround:
Esc to deactivate crop -> P to enable pointer -> Esc to exit Satty

System:
Hyprland on Arch Linux

vaknin:
Fixes #329

## Problem
Previously, the crop tool consumed all Escape keypresses even after being deactivated, preventing the global escape
handler from executing configured actions-on-escape (e.g., exit application).

## Solution
The tool now only handles Escape when the crop is active, returning `Unmodified` when inactive to allow the event to
bubble up to the global handler.

## Testing
1. Launch Satty with crop tool:  
`grim -t ppm - | ./target/release/satty --filename - --fullscreen --initial-tool=crop --actions-on-escape=exit`
3. Press Escape once → crop deactivates (handles disappear)
4. Press Escape again → application exits

## Changes
- Modified `handle_key_event()` in `src/tools/crop.rs` to check `crop.active` state before consuming Escape events

### DIFF
--- a/src/tools/crop.rs
+++ b/src/tools/crop.rs
@@ -376,11 +376,16 @@ impl Tool for CropTool {
     }
 
     fn handle_key_event(&mut self, event: KeyEventMsg) -> ToolUpdateResult {
-        if event.key == Key::Escape && self.crop.is_some() {
-            self.handle_deactivated()
-        } else {
-            ToolUpdateResult::Unmodified
+        if event.key == Key::Escape {
+            if let Some(crop) = &self.crop {
+                if crop.active {
+                    // Crop is active, deactivate it
+                    return self.handle_deactivated();
+                }
+            }
         }
+        // No crop exists or crop is inactive - let event bubble to global handler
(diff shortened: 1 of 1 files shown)


RobertMueller2:
lgtm, thanks for fixing this.

Issue #288, opened 2025-10-04
When the window is wider than the image, text is not wrapper properly in preview


RobertMueller2:
Observed in #283.

Without further investigation,  assume it's here, width is probably the window width rather than the image width. 

https://github.com/gabm/Satty/blob/6c6836f1f9a45f4b05d6167be0c34a299e630189/src/tools/text.rs#L117

<img width="1300" height="655" alt="Image" src="https://github.com/user-attachments/assets/71312cec-a8a8-4696-87d6-c1cbe041ef70" />

When saving the file, the text is properly wrapped where expected, but the bug does break WYSIWYG.

RobertMueller2:
I can no longer reproduce this in d5de7f811c6ddfccdf2cd5566b037b0cd6902991, although the wrapping does seem "weird", can't quite put my finger on it right now.

I'd say maybe this is related to pulling in femtovg updates, but it doesn't exactly match the timing of when I opened this issue.

Re-investigate after #307.

RobertMueller2:
still happening.

RobertMueller2:
Overlap with #337. 

This may actually be a feature. However the wrapping on screen needs to match what ends up in the image.

PR #336, opened 2025-11-13
feat: pan and zoom

busyoGG:
Partially based on #110, implemented pan and zoom functionality.

### Feature

* Press and hold the middle mouse button to drag and move the image.
* Use the arrow keys to move the image.
* Use the mouse wheel to zoom in when scrolling up, and zoom out when scrolling down.
* After zooming, the image size does not change with the window size. After resetting, it can resize with the window (Satty’s original behavior). At 1:1 scale, the size also does not change.

### Configuration

* `zoom-factor` : The amount of zoom per step.
* `pan-step-size`: The distance moved per arrow key press.

https://github.com/user-attachments/assets/e7b460ac-e45e-4ed5-8328-e331e9d7daaf

### Question

* Since I don’t have a touchpad, I’m unable to develop touchpad functionality.
* I have no idea how to save the content outside the original image area with the smallest possible image size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ All configuration is done either at the config file in `XDG_CONFIG_DIR/.config/s
 - `Ctrl+T`: Toggle toolbars
 - `Ctrl+Y`: Redo
 - `Ctrl+Z`: Undo
+- `Left`/`Right`/`Up`/`Down`: Pan, also available with middle mouse button drag <sup>NEXTRELEASE</sup>
 
 #### Tool Selection Shortcuts (configurable) <sup>0.20.0</sup>
 Default single-key shortcuts:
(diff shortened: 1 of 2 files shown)


RobertMueller2:
Thanks for picking this up, looks almost good to me. We need some additions to the README, but I'll take care of that.

>     * Since I don’t have a touchpad, I’m unable to develop touchpad functionality.

Yeah, that's fine, I'll open a followup feature request for this. Also for some additional shortcuts perhaps.
 
>     * I have no idea how to save the content outside the original image area with the smallest possible image size.

I know this was one of the remarks in #110. I feel we need to have a discussion about the usability of this. For starters, there should be some visual hint showing the saved image geometry. And this is going to overlap with the crop feature. Also, #337 and #288. Again, we'll discuss this in a followup. I don't want this to block this PR.

PR #340, opened 2025-11-14
fix: change Pan shortcut to avoid conflict

busyoGG:
As mentioned in the title, the default Pan shortcut conflicted with text cursor movement, so it now requires `Ctrl` as a modifier.


RobertMueller2:
I thought I had tested text, but apparently only for wrapping behaviour. Thanks for spotting this.

RobertMueller2:
Yeah we may have to adjust that again. I clearly should have taken more time while looking at the PR earlier. Ctrl+Left/Right was a word jump in the text tool. Alt is free, though.

PR #347, opened 2025-11-27
use <kbd> tag for hotkeys in readme

Axlefublr:
to make them look like hotkeys when rendered by github :3

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,13 +246,13 @@ Satty supports IME via GTK with and without preediting. Please note, at this poi
 
 You can bind a key to the following command:
 
-```bash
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Yep, that's good. Thanks :)

Can we use sh instead of bash though, please? I don't think any of the snippets are bash specific, and we already used sh in two places.

Axlefublr:
fair enough; I'm a fish user so I wasn't super sure that the code examples were fully posix compliant or not. I'll change them over to just sh

RobertMueller2:
Thanks :)

Issue #159, opened 2025-04-27
[feature] Make all changes as components so we can move around, resizing, editing text after changes


louiszn:
Satty works great! But for some problems, it's a bit annoying to redo something like typing the words again after hiting enter. I think it would be great if we can still do some modifications on what we have made to the images.

gabm:
as in .. "select any element via mouse" and then edit it again, e.g. move it around or edit text etc?

louiszn:
Yes, that's what I mean

gabm:
yep that would be great. It's on our wishlist, but it might be tricky to achieve

- first we need a hit model to check whether an existing drawable has been clicked and then activate it's tool and load the drawable - that's doable
- we need a visual model that reflects selection (usually a grey box with handles) and maybe make the handles clickable for resize etc
- then we need to add a separate undo/redo stack. Currently thats the drawable stack and only the top-most item can be "undone". Further, the undo/redo stack actions must include state changes of the widget. This could be very tricky imho 

But, it's definitely worthwhile and would set satty further apart!

Varen04:
+1
I understand the work it takes, though I do agree it would skyrocket satty even more. 

I love this project

robertwidfen:
This moves discussion from Draft PR #511 to here.

I created a branch for this now https://github.com/robertwidfen/Satty/tree/feat_edit_annotations which is based on some other smaller stuff I have not thrown into the discussion so far.

Things I want to discuss here - which are in separate commits on the branch - are:
- feat: pointer tool click raises annotation to top: 
  Is it reasonable that a click on a annotation raises it to the top?  For me it makes sense but it also changes drawing order resp. allows for changing drawing order.
- feat: Alt+Click selects overlapping annotations:
  Do we really need to be able to select annotations which are overlapping? 
- feat: (Alt+)Click starts temporary pointer tool: 
  Should a click on a annotation switch temporarily to pointer tool? I.e. there is no need to switch to pointer tool actively. I find it quite handy, but it will disallow to place a new annotation on top of an existing one, i.e. you need to place it else where and then move it.

See the video what it looks like.

https://github.com/user-attachments/assets/728e1f56-f645-4b6e-8efa-9bac662c28e9

RobertMueller2:
>     * feat: pointer tool click raises annotation to top:
>       Is it reasonable that a click on a annotation raises it to the top?  For me it makes sense but it also changes drawing order resp. allows for changing drawing order.

[...]

PR #349, opened 2025-11-27
Delete key clears the sketchboard

Axlefublr:


RobertMueller2:
pushed commit: experimental and nextrelease for the shortcut.

I think we may need Delete for #159 down the road. We can merge this for now, but we may have to reconsider the shortcut in the future.

RobertMueller2:
Noticed too late that this also happens if text tool is active. Opened bug #350.

Axlefublr:
whoops!

PR #358, opened 2025-12-20
feat: add daemon mode for faster startup                             …

rualss:
feat: add daemon mode for faster startup

      Add daemon mode that keeps GTK4 initialized between screenshot
      annotation calls, reducing window creation time from ~250ms to ~25ms.

      Features:
      - `--daemon` flag to start background daemon
      - `--show` flag to send images to running daemon
      - Per-window configuration to avoid race conditions
      - Signal handling (SIGTERM/SIGINT) for graceful shutdown
      - Unix socket IPC with length-prefixed JSON protocol
      - Security: socket permissions 0600, path validation

      Usage:
        satty --daemon &
        satty --show -f /tmp/screenshot.png --fullscreen

### DIFF
--- a/src/daemon/security.rs
+++ b/src/daemon/security.rs
@@ -1,0 +1,216 @@
+//! Security utilities for daemon mode
+//!
+//! Provides path validation and socket permission checking to prevent
+//! common security issues like path traversal attacks.
+
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
+use std::path::{Path, PathBuf};
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum SecurityError {
+    #[error("File not found: {0}")]
+    FileNotFound(PathBuf),
+
+    #[error("Path traversal detected: {0}")]
+    PathTraversal(String),
+
+    #[error("Invalid path: {0}")]
+    InvalidPath(String),
+
+    #[error("IO error: {0}")]
+    Io(#[from] std::io::Error),
+
+    #[error("Path is not a file: {0}")]
+    NotAFile(PathBuf),
+}
+
+/// Maximum path length to prevent DoS attacks
+const MAX_PATH_LENGTH: usize = 4096;
+
+/// Validate an image path for security
+///
+/// Performs the following checks (BASIC level):
+/// - Path is not empty
+/// - Path length is reasonable
+/// - File exists
+/// - Path does not contain path traversal sequences after canonicalization
+/// - Resolved path is a regular file (not a directory)
+///
+/// Symlinks are allowed in BASIC mode.
+pub fn validate_image_path(path: &str) -> Result<PathBuf, SecurityError> {
+    // Check for empty path
+    if path.is_empty() {
+        return Err(SecurityError::InvalidPath("empty path".into()));
+    }
+
+    // Check for stdin indicator (not a real path)
+    if path == "-" {
+        return Ok(PathBuf::from("-"));
+    }
+
+    // Check path length
+    if path.len() > MAX_PATH_LENGTH {
+        return Err(SecurityError::InvalidPath(format!(
+            "path too long: {} bytes (max {})",
+            path.len(),
+            MAX_PATH_LENGTH
+        )));
+    }
+
+    let path = Path::new(path);
+
+    // Check for obvious path traversal before canonicalization
+    let path_str = path.to_string_lossy();
+    if path_str.contains("..") {
+        // We'll check again after canonicalization, but catch obvious cases early
+        // This helps with error messages
+    }
+
+    // Canonicalize the path (resolves symlinks and ..)
+    let canonical = path.canonicalize().map_err(|e| {
+        if e.kind() == std::io::ErrorKind::NotFound {
+            SecurityError::FileNotFound(path.to_path_buf())
+        } else {
+            SecurityError::Io(e)
+        }
+    })?;
+
+    // After canonicalization, path should be absolute and have no .. components
+    let canonical_str = canonical.to_string_lossy();
+    if canonical_str.contains("..") {
+        return Err(SecurityError::PathTraversal(path_str.into_owned()));
+    }
+
+    // Check that it's a file, not a directory
+    let metadata = fs::metadata(&canonical)?;
+    if !metadata.is_file() {
+        return Err(SecurityError::NotAFile(canonical));
+    }
+
+    Ok(canonical)
+}
+
+/// Validate socket file permissions
+///
+/// Ensures the socket file:
+/// - Has mode 0600 (owner read/write only)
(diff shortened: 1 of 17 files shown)


RobertMueller2:
Thank you for picking this up. We'll review the PR when we get a chance, might be in the new year though.

rualss:
No problem at all. I did it because i was irritated with latency on my pc, now satty shows up instantly after taking screenshots of small areas, and i have a latency around 0.3s when i screenshot my both monitors, and i have a 4k monitor with 1.5 scaling and full hd monitor.

RobertMueller2:
Thank you for your patience. This is a comprehensive PR so it'll definitely take us a bit longer to review this. Sorry :)

If anyone else (who isn't a maintainer) feels like helping out with a review and test, by all means, please go ahead :)

Just a small thing I thought about at this stage: should we perhaps try something like `SATTY_SOCK` first before defaulting to the one in this PR? That way, users could decide how to deal with the daemon in e.g. a nested wayland session.

RobertMueller2:
I've looked through most of it and I'm ok with what I've seen so far. I still need to figure out and test daemon wide vs instance specific configs, that might require a bit more time.

I'm sticking with my remark regarding `SATTY_SOCK` above. The only thing I'd like to add right now is to use newer version of nix before we merge, but that seems entirely unproblematic.

And finally, just as a heads-up, #376 may clash regarding config use, but maybe I haven't fully understood this yet :wink:

rualss:
Yea, i think it's better to have some way to control the way daemon works with sockets. Also i think it's a good idea to add some random hash to the name of the default socket file to avoid collisions.

RobertMueller2:
Sorry for the delay in response. I'm going to add a few inline remarks. Can't promise these are going to be the last, it's tricky to find enough time to review this properly.

Overall, I have these remarks:
 - some of the tests looked to me like they were overlapping. I admit I've just had a quick glance so might be mistaken, I'll have a closer look when I get a chance. Either way, I'd prefer if they were either in tests subdirectory or in the files themselves, but not both places unless there is a reason for it?
 - I've not tried to find out why, but when I start satty without daemon mode, the toolbars are dark ish with my current theme and css. When I start it via `--show` with a daemon present, the toolbars are light. 
 - we now have many places carrying a config. This is going to require new additions to consider more places. I suppose the whole thing could do with some deduplication now. That does not necessarily have to take place in this PR, but we'll have to open a followup issue for that.
 
Lastly, I have a question. GTK itself has a mode where new instances of an app communicate to an existing instance via DBus, this could in principle also be suitable for a daemon mode. There may be additional challenges with this, and perhaps the dependency on working  DBus itself for the daemon mode is something we wouldn't want at all. But out of curiosity, have you considered it, and if so, why have you decided against it?

src/daemon/security.rs
```diff
@@ -0,0 +1,211 @@
...
+        if e.kind() == std::io::ErrorKind::NotFound {
+            SecurityError::FileNotFound(path.to_path_buf())
+        } else {
+            SecurityError::Io(e)
+        }
+    })?;
+
+    // After canonicalization, path should be absolute and have no .. components
+    let canonical_str = canonical.to_string_lossy();
+    if canonical_str.contains("..") {
```
RobertMueller2: This shouldn't happen after canonicalising a path. Unless you have a demonstration that shows otherwise, I suggest to remove it along with the error kind above. I think it's not needed because it's never going to be triggered.

There is socket security by controlling unix permissions on the same. If anyone is concerned about that not being enough, there are things like bwrap or firejail to limit what the daemon could access.

cybergrind:
It is leaking GPU memory, so I wouldn't recommend to merge it until leak is fixed
This is fixes it for me, but might require more careful analysis

```
diff --git a/src/main.rs b/src/main.rs
index 3a8f8ad..ce0b605 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -471,6 +471,16 @@ fn run_daemon() -> Result<()> {
     // Remove stale socket if any
     remove_stale_socket()?;
 
+    // GTK4's GskVulkanRenderer leaks per-window swapchain memory on at least
+    // NVIDIA — every annotation window we open and close grows GPU memory by
+    // ~20 MiB even though the App, SketchBoard, FemtoVGArea, GdkSurface and
+    // gtk::Window all drop. The GL renderer reclaims correctly. Honor an
+    // explicit user override; otherwise force GL for daemon mode.
+    if std::env::var_os("GSK_RENDERER").is_none() {
+        // SAFETY: single-threaded at this point, before GTK init.
+        unsafe { std::env::set_var("GSK_RENDERER", "gl") };
+    }
+
     // Load OpenGL
     load_gl()?;
     generate_profile_output!("daemon: loaded gl");
```

RobertMueller2:
> It is leaking GPU memory, so I wouldn't recommend to merge it until leak is fixed This is fixes it for me, but might require more careful analysis
> 
> ```
> diff --git a/src/main.rs b/src/main.rs
> index 3a8f8ad..ce0b605 100644
> --- a/src/main.rs
> +++ b/src/main.rs
> @@ -471,6 +471,16 @@ fn run_daemon() -> Result<()> {
>      // Remove stale socket if any
>      remove_stale_socket()?;
>  
> +    // GTK4's GskVulkanRenderer leaks per-window swapchain memory on at least
> +    // NVIDIA — every annotation window we open and close grows GPU memory by
> +    // ~20 MiB even though the App, SketchBoard, FemtoVGArea, GdkSurface and
> +    // gtk::Window all drop. The GL renderer reclaims correctly. Honor an
> +    // explicit user override; otherwise force GL for daemon mode.
> +    if std::env::var_os("GSK_RENDERER").is_none() {
> +        // SAFETY: single-threaded at this point, before GTK init.
> +        unsafe { std::env::set_var("GSK_RENDERER", "gl") };
> +    }
> +
>      // Load OpenGL
>      load_gl()?;
>      generate_profile_output!("daemon: loaded gl");
> ```

I haven't spent much effort to educate myself on this, but as far as I understand, this is a GTK issue related to certain hardware. The underlying cause doesn't have much -if any- observable consequence without daemon mode, but in principle it is already present without it. The daemon mode happens to aggravate it to a noticeable magnitude by virtue of persisting for potentially a long time.

We're not a distro or a desktop env, and I don't think an application should set env vars like this. We can add this to documentation and/or FAQ, perhaps even print out a warning under certain circumstances. But actually deciding which setting is appropriate can be hardware and/or driver dependent and therefore should be done by the user.

cybergrind:
@RobertMueller2 i think that satty as it is right now - fine, even if we have leak here - it doesn't survive for a long.

but if you want to have daemon mode enabled - this might become an issue and very quickly
and I'm 100% not sure if this is a correct way to handle it or it is very specific to some wayland + gtk internals

Issue #366, opened 2025-12-30
release.nu script is broken


martintrojer:
### Version

0.2.0

### Description

2 issues.

1/ Cargo.toml parsing is wrong
2/ command_line.rs has moved

RobertMueller2:
Fixed by #367

PR #367, opened 2025-12-30
fix: release.nu 

martintrojer:
fixes issues mentioned in #366


martintrojer:
lint errors also present on main branch

RobertMueller2:
I suppose we would have noticed with next release. Thanks for catching this before us ;)

Issue #371, opened 2026-01-01
Change the text origin while editing text


lazyWombat:
Hi, I love the utility and it works well, but there is just one thing that requires more time than I like: text adjustments. At the moment I pretty much just undo my last action, pick the new location and then paste the text into the edit box. It is a bit annoying especially if I want to align it on the right side.

I saw #159 and while it will solve the problem it looks like it requires a bit of effort. In the meantime, if possible, could you guys implement some shortcut keys (like Alt+arrows) to move the origin of the text while editing it? Or Ctrl+Click to set the new origin point. It will save a lot of time.

RobertMueller2:
Thanks for the issue and related PR.

Setting new origin via mouse also sounds good, let's keep this issue open for this. To be in line with the suggestion I made in said PR, I'd propose Ctrl+Alt+click to trigger it.. That said, maybe we should also consider dragging if the starting point is within the text box. Without confirming my assumption: I believe we should have pretty much everything to tell whether the click location is within the bounds of the current text edit.

robertwidfen:
Fixed by #607

PR #372, opened 2026-01-02
feat: text movement

lazyWombat:
This PR adds the ability to move and nudge text using arrow keys. See #371.

### Feature

- Use ALT and the arrow keys to move the text (the distance is configurable, 50 pixels by default).
- Use ALT+SHIFT and the arrow keys to nudge the text (by 1 pixel).

### Configuration

- text-move-length: The distance to move the text per arrow key press.

### DIFF
--- a/src/tools/text.rs
+++ b/src/tools/text.rs
@@ -874,12 +874,22 @@ impl Tool for TextTool {
                                 Self::handle_text_buffer_action(t, Action::Select, other_mask);
                         }
                         m if m == ModifierType::ALT_MASK | ModifierType::CONTROL_MASK => {
-                            tool_update_result =
-                                Self::handle_text_buffer_action(t, Action::MoveOrigin, ctrl_alt_mask);
+                            tool_update_result = Self::handle_text_buffer_action(
+                                t,
+                                Action::MoveOrigin,
+                                ctrl_alt_mask,
+                            );
                         }
-                        m if m == ModifierType::ALT_MASK | ModifierType::CONTROL_MASK | ModifierType::SHIFT_MASK => {
-                            tool_update_result =
-                                Self::handle_text_buffer_action(t, Action::NudgeOrigin, ctrl_alt_mask);
+                        m if m
+                            == ModifierType::ALT_MASK
+                                | ModifierType::CONTROL_MASK
+                                | ModifierType::SHIFT_MASK =>
+                        {
+                            tool_update_result = Self::handle_text_buffer_action(
+                                t,
+                                Action::NudgeOrigin,
+                                ctrl_alt_mask,
+                            );
                         }
                         m if m == ModifierType::CONTROL_MASK | ModifierType::SHIFT_MASK => {
                             tool_update_result =
@@ -1321,7 +1331,7 @@ enum Action {
     MoveCursor,
     Select,
     MoveOrigin,
-    NudgeOrigin, 
+    NudgeOrigin,
 }
 
 impl TextTool {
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Thanks, this is a nice addition.

I do have one concern: this change would override panning shortcuts. I'm saying "would" because I just noticed we broke the functionality from within text tool somewhere. I'll record a bug for that, because IMO panning from within text editing should work as well.

That said, there isn't a great deal of possibilities. Just Ctrl and Ctrl+Shift are kinda reserved for word jump. The best I could come up with is Ctrl+Alt+Cursor for moving and Ctrl+Alt+Shift+Cursor for nudging. What do you think?

lazyWombat:
Hi, I'm new here so I would be happy with Ctrl+Alt and Ctrl+Alt+Shift shortcuts. Pretty much as long as I can adjust the text I will be happy.

Having said that, though, I think panning as a global shortcut is a bit too much. I think it will limit the future extensibility. For example, I thought that the Ctrl modifier will later align the text with other objects (if/when Satty will remember them). I was a bit too lazy to implement it now since it requires to draw some guides on the screen, but even now it is possible to have some sticky centered text. So if it will be me, I would make pan as a tool with its own shortcuts and let other tools have their own.

At the end though, I'm quite happy with either solution and I strongly suspect that implementing #159 will introduce quite a few breaking changes, so shortcuts will be a minor thing in comparison.

Ah, one more thing that I really like to add is the shadow for the text. A lot of my usage involve annotating some screenshots sometimes I struggle to pick a color when the background is complicated. A differently colored shadow (or outline) will help with that. But again it is probably a bigger feature since circles and rectangles can benefit from it as well.

lazyWombat:
I've updated the keyboard shortcuts in the PR but I'm not quite sure how to fix the pan shortcuts neatly.

RobertMueller2:
> I've updated the keyboard shortcuts in the PR but I'm not quite sure how to fix the pan shortcuts neatly.

That's fine. I mentioned them because of the clash, but we have to fix them in a separate issue/PR. It wasn't my intention to add that task to your PR. ;)

RobertMueller2:
> Hi, I'm new here so I would be happy with Ctrl+Alt and Ctrl+Alt+Shift shortcuts. Pretty much as long as I can adjust the text I will be happy.
> 
> Having said that, though, I think panning as a global shortcut is a bit too much. I think it will limit the future extensibility. For example, I thought that the Ctrl modifier will later align the text with other objects (if/when Satty will remember them). I was a bit too lazy to implement it now since it requires to draw some guides on the screen, but even now it is possible to have some sticky centered text. So if it will be me, I would make pan as a tool with its own shortcuts and let other tools have their own.

That's a fair point. This might not survive the long term ideas, or in the long run we may need a strategy to explicitly target a global or tool specific shortcut, and ideally without adding confusion or a lot of complexity. :wink: But who knows when these things will happen, in the meantime there are smaller incremental changes like this one. And with those I feel we shouldn't mask existing shortcuts (with Enter and Esc already being necessary but hopefully intuitive exceptions), as well as try to avoid breaking existing functionality. That's why I had to point out there is a clash here (if we fix bug #374).

> Ah, one more thing that I really like to add is the shadow for the text. A lot of my usage involve annotating some screenshots sometimes I struggle to pick a color when the background is complicated. A differently colored shadow (or outline) will help with that. But again it is probably a bigger feature since circles and rectangles can benefit from it as well.

Hmm, I suppose this could work similar to to the fill shapes button. That part is probably not complicated. The shadow itself may be a separate path with an offset, but I haven't attempted this. That part may actually be more complicated. :) Sure, feel free to open a new feature request issue for this.

RobertMueller2:
I couldn't get nudging to work now. I think we shouldn't use `m.contains()` but actually `m == `.

And fixing the Panning stuff is actually easy, just need a separate match arm for `ALT_MASK` to return `ToolUpdateResult::Unmodified`. I'll quickly add a fix for this and formatting.

RobertMueller2:
pushed commits: fixing nudging, panning and formatting.

RobertMueller2:
LGTM, thanks :)

Issue #377, opened 2026-01-03
actions-on-right-click happening too often


RobertMueller2:
### Version

recent git

### Description

sketchboard has a number of places where it's using actions_on_right_click. I think this is not correct in many places (e.g. scrolling) and should really only happen if there is an actual right click.

This probably happened a while ago, haven't investigated.

PR #378, opened 2026-01-03
fix: right click actions triggered too often

RobertMueller2:
Closes: #377


gabm:
Is request_render ever being used with a callback , or is this callback always empty now? Maybe we can clean up?

RobertMueller2:
> Is request_render ever being used with a callback , or is this callback always empty now? Maybe we can clean up?

Yes, there are still several occurrences with non-empty parameter.

PR #385, opened 2026-01-05
feat!: change early-exit to allow individual triggers

### LINKED ISSUE #152: Early Exit only on enter action
I have two workflows when it comes to taking screenshots. I want to:

1. Hit enter and close `satty` and save the file to `~/Pictures/Screenshots/$timestamp.png` (which is done)
2. Click on Copy and manually close it if needed.

The `--early-exit` flag exits on both copy and save. I wish this could be configured.

RobertMueller2:
Closes: #152

This allows the following configs

```
early-exit = true
early-exit = [ "all" ]
early-exit = [ "copy", "save", "save-as" ]
```
as well as command line
```
--early-exit
--early-exit=all
--early-exit=copy,save,save-as
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,10 +110,12 @@ resize = { mode = "size", width=2000, height=800 }
 floating-hack = true
 # Change to true to automatically copy to clipboard after every annotation change
 auto-copy = false
-# Exit directly after copy/save action. 0.20.1: Does not apply to save as
-early-exit = true
-# Exit directly after save as (0.20.1)
-early-exit-save-as = true
+# Exit directly after copy/save action. NEXTRELEASE: change to list of triggers
+# Note that exit-early-save-as was removed with NEXTRELEASE.
+early-exit = ["all"]
+# is equivalent to both
+# early-exit = ["copy", "save", "save-as"]
+# early-exit = true
 # Draw corners of rectangles round if the value is greater than 0 (0 disables rounded corners)
 corner-roundness = 12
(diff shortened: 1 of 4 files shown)


gabm:
My mind model here was: 

- we have early exit (press escape/enter to close)
- we have actions on exit, that could include save, copy etc.. that would allow for "save, save as and copy" in sequence

Would the solution not be to just ensure that all actions are properly run before exiting and therefore only keeping one single "early exit flag"?

README.md
```diff
@@ -308,7 +308,7 @@ Options:
       --title <TITLE>
           Experimental feature (NEXTRELEASE): Set window title
       --app-id <APP_ID>
-          Experimental feature (NEXTRELEASE): Set toplevel app_id. Note that this applies gtk format expectations
```
robertwidfen, line 311: This likely should have been part of #400 - but I am fine to merge it here.
Splitting and moving around changes is harder in git than jj - yeah I know, I'm promoting it. 😇

robertwidfen (approved):
LGTM - have tested it - was not even aware that there was/is an option for this and happy with the default "true".

README.md
RobertMueller2: Yeah, I changed the code comment later on and didn't update the readme. Funny though, I could have sworn we had an action that checks `make update-readme`, which would have prevented this. But apparently I'm imagining things.

And then when I spotted it it happend when I ran update-readme, i didn't feel like extracting that. I know we have a rule in Contrbuting.md that asks not to, but it was easier to break it. :wink:

PR #399, opened 2026-01-26
fix: stale scale factor

### LINKED ISSUE #293: Drawing first element has wrong position in multi monitor with different dpi-scaling
After taking a screenshot the first element I'm drawing starts at the wrong position. The second annotation draws in the right position (mouse position).

I can reproduce the problem when there is a scale factor on one of the monitors. When I set this to the same value as the other monitor there is not offset.

To reproduce I found the following settings in `hyprland.conf`.

does not work:

    monitor=eDP-1,preferred,auto,1.6

this works:

    monitor=eDP-1,preferred,auto,1

When I move the window to the other monitor then it also works with the scale factor.

RobertMueller2:
Closes: #293

When starting satty e.g. on a DPI scale factor = 1.6 screen and moving the window to a scale factor = 1.0 screen, the scale factor of femtovg area is stale for the first tool used.

This causes the first tool after window move to appear in a wrong position.

Fix this by adding an event handler for changed scale-factor.

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -37,6 +37,7 @@ pub enum SketchBoardInput {
     CommitEvent(TextEventMsg),
     Refresh,
     Exit,
+    ScaleFactorChanged,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


src/sketch_board.rs
```diff
@@ -1116,6 +1117,10 @@ impl Component for SketchBoard {
                 self.handle_exit();
                 ToolUpdateResult::Unmodified
             }
+            SketchBoardInput::ScaleFactorChanged => {
+                self.renderer.resize(0, 0);
```
gabm: this is 0,0?

PR #468, opened 2026-03-27
[needs a human pls] add pixelate functionalty (on alt) to blur

kikeijuu:
Detailed more in https://github.com/Satty-org/Satty/issues/90 
Inspired by flameshot's pixelation but much less secure as I have the skills of a Victorian chimney sweep >.> 

Written largely with (local) AI as I don't really know rust, but tested by 3 different people now and builds without errors ^^ Been chatting with several other satty users on a fluxer and we love the tool, and as they enjoy this pixelation-on-alt so I'm making this PR sooner rather than later. I understand however if that prevents this from being merged!

Various images with both blur and pixelate for comparison:

image

image

image

### DIFF
--- a/src/tools/blur.rs
+++ b/src/tools/blur.rs
@@ -1,9 +1,9 @@
 use std::cell::RefCell;
 
 use anyhow::Result;
-use femtovg::{Color, ImageFilter, ImageFlags, ImageId, Paint, Path, imgref::Img};
+use femtovg::{Color, ImageFilter, ImageFlags, ImageId, Paint, Path, imgref::Img, rgb::Rgba};
 
-use relm4::{Sender, gtk::gdk::Key};
+use relm4::{Sender, gtk::gdk::{Key, ModifierType}};
 
 use crate::{
     configuration::APP_CONFIG,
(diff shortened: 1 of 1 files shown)


kikeijuu:
> How is this suggested approach supposed to behave when I select 100% of the area, i.e., when there's no "fringe" left? This may be tricky to do with flameshot, but it's definitely a possibility in Satty. Even if it makes no sense, it is a valid input. The algorithm is guarded against it, I just have no idea what the result would look like. I assume if this feels like a limitation, we can document it in readme or later on, documentation.

image

Looks like this! as you can see, it makes a rectangle to fill what's there, even when I drag over it entirely. It rounds the corners too, as I haven't changed that in my config

> Depending on how close the implementation is, we may have to add a credit note or even figure out how/if it can be re-licenced

Since this isn't actually nearly as secure I don't think we will be good but I don't know much about licenses ^^

> Now I am thinking that pixelation may be a different tool from blur, perhaps we want both a standard pixelation and the fringe pixelation as a pixelate tool, while leaving the blur tool where it is. Or we can leave that part for a followup change if anyone's bothered.

I'm curious what the og issue maker's take is on this, but I have no opinions either way. I'm not very good with design, but I'm happy as long as everything is kb accessible :D

kikeijuu:
> we can keep the PR open for reference purposes for now/until we have an implementation. I mean, if people want to use it as a patch for their own installation, if they build themselves anyway, that might be useful.

keeping open as per this. I'll keep this synced until a human comes along :D

PR #444, opened 2026-03-02
feat: indicate active editing in toolbar

RobertMueller2:
This adds a highlight (using adwaita @accent_color) for the button for a tool that is actively editing. This works for:

     - arrow
     - blur
     - brush
     - crop
     - ellipse
     - highlight
     - line
     - rectangle
     - text (already has everything that's needed)

This is mostly an indicator for tools that have states like crop or text, but can also work for the others while drawing.

editing
satty-20260302-185854

not editing
satty-20260302-185903

PR #469, opened 2026-03-27
feat: add pixelate and pseudo-pixelate

### LINKED ISSUE #90: [feature] low res pixel blur
At the moment there is only a smooth blur—a lot of data can be retrieved from smooth blurs, like text and decent approximations of images. The default blur seems pretty heavy, so maybe this is unneeded, but I think it would be really helpful to have a pixel blur which can be applied as a distinct option from the existing smooth (Gaussian?) blur, for censoring names ect.

Spectacle is an example of a screenshot tool that I know has both blur types.

Thanks for an awesome tool!

RobertMueller2:
Closes: #90

### TL;DR: What's the holdup?

~2026-09-03
I hit a snag. This was initially copied from blur and thus inherited its shortcomings, e.g. #602. Blur, like this, needs to work on the underlying image, rather than using canvas to create a screenshot of it. At least, we need to create the groundwork in order to allow both, right now `draw()` just has the canvas, not the image.~

### in case you actually do want to read

This implements a pixelate tool with two modes
- regular mode with a normal block pixelation, this is, depending on selected area and image resolution, reversible
- fringe inpaint+pixelate that takes data from _outside_ the selected area and applies distance weighting. The original area is entirely overwritten by the interpolated data. Then, the regular pixelation function creates a block-pixelate look from this.

Also adds "fringe inpaint" as a standalone tool because it was easy to add that.

Dragging the rect to outside of the visible area could cause panics, this was because `canvas.screenshot()` only contains the viewport. Now for inpainting, you'd still kinda want to use data from outside of the viewport, and to a lesser degree this applies for the pixelation algorithm. We address this by passing an image to new signature `draw_baselayer`.

Tasks
- [x] figure out a better icon - help needed please. for both pixelate and pseudo-pixelate
- [x] reach out to flameshot guys regarding credits/licence
- [x] double check if south and east buffers aren't actually off by one
- [x] figure out if BLOCKSIZE is good at a value of 32. This may actually need to be adjustable. Follows size+annotation-size-factor now. Annotation-size-factor gets rounded to 1, S is 4, M is 8, L is 16. These values may need some fine tuning in the future.
- [x] find a user friendly solution for "selected area too small to pixelate". Shows different colours for rectangle now. But perhaps needs more conditions, e.g. when there's no fringe, the pixelation can fail as well. Perhaps use different cursors here.
- [x] this user friendly solution must also be applied to resizing/moving via pointer
- [x] alt is probably not a good way to switch between pixelate and pseudo-pixelate, it's inconsistent with other tools. implement as variant mode on top of #600. Now removed in favour of distinct tool entries.
- [x] indicate editing consistent with #444
- [x] rework on top of #607 - rebased now but we need to consider the layer we draw on and make this movable
- [x] address out of sight areas
- [x] rework fringe to use the available fringes
- [x] if fringes are missing, use the other ones. also, only use intersection of image and selection bounds.

Known limitations:
- could use some noise, leave that for followup
- stacking several instances of the drawables introduced here will not really stack but always apply onto the base image. We'll leave this for #602 to fix.

This PR was inspired by
- https://github.com/flameshot-org/flameshot/pull/3765/changes
- #468 by kikeijuu (superseded by this PR)

Regarding the credits, the idea with taking "fringes" was taken from said PR, but I reimplemented the gist of it according to my understanding. I did, for example, skip any additional noise which might be suitable for adding later. Also, actually iterating the source (fringes) and target data (overwritten area) works a bit different due to the objects used. I don't think we need to talk about a separate licence for this bit, but we'll see what the flameshot guys say about this. This is being discussed in https://github.com/flameshot-org/flameshot/issues/4926

https://github.com/user-attachments/assets/77d167ba-ac39-4d4b-b6cc-fd7edec8ac0a

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Default single-key shortcuts:
 - <kbd>t</kbd>: Text tool
 - <kbd>m</kbd>: Numbered Marker tool
 - <kbd>u</kbd>: Blur tool
+- <kbd>x</kbd>: Pixelate tool
 - <kbd>g</kbd>: Highlight tool
 
 ### Tool Modifiers and Keys
@@ -79,6 +80,7 @@ Default single-key shortcuts:
 - Highlight: Hold <kbd>Ctrl</kbd> to switch between block and freehand mode (default configurable, see below), hold <kbd>Shift</kbd> for a square (if the default mode is block) or a straight line (if the default mode is freehand)
 - Line: Hold <kbd>Shift</kbd> to make line snap to 15° steps
 - Rectangle: Hold <kbd>Alt</kbd> to center the rectangle around origin, hold <kbd>Shift</kbd> for a square
+- Pixelate: Hold <kbd>Alt</kbd> to use pixelation that takes data from outside the selection as a source. <sup>NEXTRELEASE</sup>
 - Text:
   - Press <kbd>Shift+Enter</kbd> to insert line break.
   - Combine <kbd>Ctrl</kbd> with <kbd>Left</kbd> or <kbd>Right</kbd> for word jump or <kbd>Ctrl</kbd> with <kbd>Backspace</kbd> or <kbd>Delete</kbd> for word delete.
@@ -120,7 +122,7 @@ early-exit = true
 early-exit-save-as = true
 # Draw corners of rectangles round if the value is greater than 0 (0 disables rounded corners)
 corner-roundness = 12
-# Select the tool on startup [possible values: pointer, crop, line, arrow, rectangle, text, marker, blur, brush]
+# Select the tool on startup [possible values: pointer, crop, line, arrow, rectangle, text, marker, blur, pixelate, brush]
 initial-tool = "brush"
 # Configure the command to be called on copy, for example `wl-copy`
 copy-command = "wl-copy"
@@ -189,6 +191,8 @@ ellipse = "e"
 text = "t"
 marker = "m"
 blur = "u"
+# NEXTRELEASE
+pixelate = "x"
 highlight = "g"
 
 # Font to use for text annotations
@@ -266,7 +270,7 @@ Options:
       --corner-roundness <CORNER_ROUNDNESS>
           Draw corners of rectangles round if the value is greater than 0 (Defaults to 12) (0 disables rounded corners)
       --initial-tool <TOOL>
-          Select the tool on startup [aliases: --init-tool] [possible values: pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, highlight, brush]
+          Select the tool on startup [aliases: --init-tool] [possible values: pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, pixelate, highlight, brush]
       --copy-command <COPY_COMMAND>
           Configure the command to be called on copy, for example `wl-copy`
       --annotation-size-factor <ANNOTATION_SIZE_FACTOR>
@@ -444,3 +448,8 @@ Made with [contrib.rocks](https://contrib.rocks).
 The source code is released under the MPL-2.0 license.
 
 The Font 'Roboto Regular' from Google is released under Apache-2.0 license.
+
+## Credits
+
+- Pixelate "independent mode" was inspired by https://github.com/flameshot-org/flameshot/pull/3765/changes
+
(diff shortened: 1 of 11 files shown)


RobertMueller2:
for brighter images, the "independent mode" result is a bit too bright. Maybe the weighting math isn't great yet.

EDIT: need to divide everything by two if I have a a horizontal and a vertical influence, they each add to 1. that looks a lot better. Unfortunately, now I have to redo the recording.

RobertMueller2:
Perhaps we also really need to call this "pseudo-pixelate", no need to re-invent lingo? :wink:

robertwidfen:
Regarding the icon, maybe "checkerboard"?

satty-pixelate-icon-checkerboard

But I like "eye-off-regular" better as it fits more into the existing icons.

satty-pixelate-icon-eye-off-regular

Also I see a small issue if the pixelate region is a the border and covers less than one "pixel" in width/height it will not pixelate anything.

satty-pixelate-issue-at-border

RobertMueller2:
> Also I see a small issue if the pixelate region is a the border and covers less than one "pixel" in width/height it will not pixelate anything.

yup, I noticed that too while using/testing this locally, at the very least it needs some indication whether the selected area is big enough or not. That'll need some rework, i'll place this in draft mode for now.

pointydev:
Hey there, just came across this pull and noticed the following in the OP:
> - [ ] figure out if BLOCKSIZE is good at a value of 32. This may actually need to be adjustable.

On Flameshot this is handled using the scrollwheel to change tool size (also the case for the rest of the sized tools, such as line width, etc). I personally use a tool size of 2 on most occasions, equating to 6px blocks for me:



Adjustable block size would be highly appreciated as it can help emphasise that an area is meant to be pixelated in a final image, depending on the scale of the surrounding elements.

robertwidfen:
Thank you for sharing this. Personally I do not switch the tool size much, but I like the idea for blur and pixelate, there the S-M-L might not be sufficient. Pixelate so far does not take S-M-L into account.

I was also thinking about integrating pixelate into the existing blur tool to avoid adding another toolbar button.  But currently we have no concept of tool individual styles, which might be handy to switch between blur/pixelate, add double arrow, switch rounded rectangle on/off, etc.

RobertMueller2:
I would tie it to S-M-L and annotation size factor, and a static modifier, which is from top of my head what the other tools do. I wouldn't add a scroll wheel logic specifically for this -- but the scroll wheel does work on the annotation size, so at least there's that. ;)

RobertMueller2:
> I was also thinking about integrating pixelate into the existing blur tool to avoid adding another toolbar button. But currently we have no concept of tool individual styles, which might be handy to switch between blur/pixelate, add double arrow, switch rounded rectangle on/off, etc.

If such a concept was ready to use, I might add this to the blur tool. But actually coming up with a concept and discussing it, I think we should open a separate issue :)

README.md
```diff
@@ -98,6 +98,7 @@ Default single-key shortcuts:
 - <kbd>t</kbd>: Text tool
 - <kbd>m</kbd>: Numbered Marker tool
 - <kbd>u</kbd>: Blur tool
+- <kbd>x</kbd>: Pixelate tool
```
robertwidfen, line 101: Fringe

src/tools/pixelate.rs
```diff
@@ -0,0 +1,566 @@
...
+        size: Vec2D,
+    ) -> Result<Option<ImageId>> {
+        let transformed_pos = canvas.transform().transform_point(pos.x, pos.y);
+        let average_scale = canvas.transform().average_scale();
+        let transformed_size = size * average_scale;
+
+        let blocksize = self
+            .style
+            .size
+            .to_blocksize(self.style.annotation_size_factor);
```
robertwidfen, line 105: .min(size.x.min(size.y) as usize);

Instead of failing I would prefer to make it work always.
Then the check "big_enough" below can be removed.

robertwidfen (changes requested):
Cool stuff

README.md
RobertMueller2: yeah. hmm .. i suppose it should be mentioned. does it need a default shortcut?

src/tools/pixelate.rs
RobertMueller2: I'm not sure I agree. Your suggestion works, of course. But it can easily lead to very different applied block sizes.

satty-20260903-222314

It might be tricky for the user to figure out which size to drag to achieve uniform results between different areas. I'd rather tell them "this doesn't work yet" via some UI feedback than trying to do something.

README.md
RobertMueller2: oh, i suppose i got you wrong the first time. fixed ;)

robertwidfen:
lgtm - have reviewed the code and tested it.

With the remainder I still have some odd feelings and actually I am preparing a fix for review - but you probably know better than anyone how annoying all this pixel-pushing and its edge cases can be ... 😇

RobertMueller2:
Sure, happy about further improvements. :) but also good to get this merged finally :D

PR #479, opened 2026-04-05
fix: replace println! with eprintln!

### LINKED ISSUE #478: inconsistent println! and eprintln!
### Version

1059c967ed217008ad2a873368c7041a392420dd

### Description

We may write image to stdout, so IMO we should never place any other messages (mostly error messages) on stdout. But there are several places that use `println!` instead of `eprintln!`.

```
> grep -r "\<println" src | wc -l
11
```

RobertMueller2:
Closes: #478


robertwidfen (approved):
Looks good to me.

Issue #481, opened 2026-04-11
feat: Add Undo as an Action variant + double-click to trigger actions-on-enter


clansty:
### 1. `Undo` as an `Action` variant

Currently the `Action` enum supports `save-to-clipboard`, `save-to-file`, `save-to-file-as`, `copy-filepath-to-clipboard`, and `exit`. These can be composed in `actions-on-enter`, `actions-on-escape`, and `actions-on-right-click`.

I'd like to propose adding `undo` as a new `Action` variant. This would allow configurations like:

```
--actions-on-right-click undo,exit
```

**Semantics**: `undo` is a fallible action — if it succeeds (something was undone), stop processing further actions in the list. If it fails (nothing to undo), continue to the next action. This makes `undo,exit` mean "undo one step; if nothing left to undo, exit".

**Use case**: Right-click-to-undo is a common interaction pattern in annotation/drawing tools. Combined with `exit` as a fallback, it gives a natural "undo or cancel" flow without needing a dedicated config flag.

### 2. Double-click to trigger configurable actions

Add `--actions-on-double-click` as a new configurable action list, consistent with the existing `actions-on-enter`, `actions-on-escape`, and `actions-on-right-click` options.

When `actions-on-double-click` is not set (empty), it falls back to `actions-on-enter` behavior for backwards compatibility.

Double-click is only processed when the active tool does not stop propagation, so it does not interfere with existing double-click behavior (e.g. text tool word selection).

**Use case**: Double-click to confirm/copy is a natural gesture for screenshot crop workflows.

RobertMueller2:
> **Semantics**: `undo` is a fallible action — if it succeeds (something was undone), stop processing further actions in the list. If it fails (nothing to undo), continue to the next action. This makes `undo,exit` mean "undo one step; if nothing left to undo, exit".

I don't find this very intuitive. If action chains had any control flow, which may be something to consider in #89, I could see fallible actions. But it would have to be something explicit along the chain, not implicit in the action.

FWIW, undo likely is by an order of magnitude more fallible than save or copy.
 
> **Use case**: Right-click-to-undo is a common interaction pattern in annotation/drawing tools.

Do you have any examples for this, please?

> Combined with `exit` as a fallback, it gives a natural "undo or cancel" flow without needing a dedicated config flag.

Er... I do not find this natural at all, sorry.

[...]

Issue #450, opened 2026-03-08
More general/unified approach to keybindings


camoz:
### Description

I just tried to add a keybinding for `q` to exit satty, but apparently this is not possible. Also I noticed that keybindings are currently defined in two places: the `[keybinds]` section and the `[general]` section (namely the `actions-on-*` settings). 

So I thought it would be nice to unify the approach to keybindings and have them only defined in the `[keybinds]` section, e.g. by adding `Enter`, `LeftMouse`, `RightMouse`, etc as possible "keys", and `exit`, `save`, etc as possible "actions". AFAICS this should be possible in a backwards compatible manner.

I think it would also be nice to have the mappings defined as `keybind = action` instead of `action = keybind`, as generally one keybind can only have one action, but one action can have multiple keybinds associated with it (e.g. `Esc` and `q` could both be mapped to `exit`). But this is probably not doable in a backwards compatible manner and not so important I think.

RobertMueller2:
In the long run, I think shortcuts should be bindable to actions/action chains or tools or colours etc.  We kinda need to do this for #89. Backwards compatibility is important, though.

robertwidfen:
A first version is here: https://github.com/robertwidfen/Satty/tree/feat_generic_key_bindings

Please comment & test ;-)

The description of the change is below
----

Changes syntax for [keybinds] section in config file to 

    "key" = "tool-or-command"

Also adds:
- commands for size selection
- automatic hint generation for all buttons

Removes:
- layout independent key binding - instead bind what you warnings
- single letter binding restriction 

Old style is still supported but will generate deprecation warnings.

The old code always sent key event into the IME handling before
they were processed as key event. Thus there were two places for
bindings, bindings with a modifier and single letter bindings.
But IME is for text input, thus it is now only activated for
the text tool.

RobertMueller2:
Well done, and looks sound at first glance, and it's probably advanced enough to open a PR, if you ask me. Haven't had any opportunity to test yet, might take a few days.

One remark regarding the actions for enter and escape, I think in the future we may want to have something like named action chains (on the back of #89). E.g. something like actionchain1 which runs a shell command and then quits. For that, we'd probably want to refer to them in a dynamic way in the shortcuts, e.g. instead of
`"<Control>q" = "run-actions-on-escape"`
something like
`"<Control>q" = "action-chain:escape-actions"`

[...]

PR #487, opened 2026-04-14
feat: add Undo and UndoOrExit action variants

### LINKED ISSUE #485: Add Undo and UndoOrExit action variants for action chains
## Description

Add `undo` and `undo-or-exit` as new action variants that can be used in action chains (`actions-on-enter`, `actions-on-escape`, `actions-on-right-click`, etc.).

- **`undo`**: Plain action that undoes one step. No-op if nothing to undo, continues the chain regardless.
- **`undo-or-exit`**: Compound action — undoes one step if possible; if nothing to undo, exits the application. This enables the common "right-click to undo, or exit when done" workflow used by tools like Snipaste and PixPin.

### Use case
Right-click-to-undo is a common interaction pattern in annotation/drawing tools. Combined with `undo-or-exit`, users can configure `--actions-on-right-click undo-or-exit` for a natural "undo or cancel" flow.

### Implementation notes
- A new `process_actions` method handles Undo/UndoOrExit before passing remaining actions to the render pipeline.
- Enter/Escape/right-click are routed through `process_actions` so Undo works consistently in all action chains.
- No implicit control flow — the compound behavior is explicit in the `undo-or-exit` action name.

Related discussion: #481

clansty:
## Summary

- Add `undo` and `undo-or-exit` as new action variants for action chains
- `undo`: plain action, no-op if nothing to undo, always continues the chain
- `undo-or-exit`: compound action — undo one step if possible, exit if nothing to undo
- Route Enter/Escape/right-click through new `process_actions` method for consistent Undo support
- No implicit control flow — compound behavior is explicit in the action name

Closes #485
Related discussion: #481

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,13 +142,13 @@ primary-highlighter = "block"
 # Disable notifications
 disable-notifications = false
 # Actions to trigger on right click (order is important)
-# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit]
+# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit, undo, undo-or-exit]
 actions-on-right-click = []
 # Actions to trigger on Enter key (order is important)
-# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit]
+# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit, undo, undo-or-exit]
 actions-on-enter = ["save-to-clipboard"]
 # Actions to trigger on Escape key (order is important)
-# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit]
+# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit, undo, undo-or-exit]
 actions-on-escape = ["exit"]
 # Action to perform when the Enter key is pressed [possible values: save-to-clipboard, save-to-file]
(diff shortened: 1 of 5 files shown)


RobertMueller2:
Apologies for the delay in response, and thanks for the PRs.

`process_actions` can toss around the order. E.g. if I use `--actions-on-right-click=copy-filepath-to-clipboard,undo-or-exit` then `undo-or-exit` happens first, and if it exits, no filepath is copied.

I understand that it might be tricky to trigger a redraw for the undo, but we need to find a way to ensure the order of the actions. Ideally, we don't have action processing in two places. Perhaps some better queueing of the actions while escaping the action processing temporarily, from top of my head save-to-file-as may serve as an example and could be done more generically?

Also, my gut feeling is that I'd prefer not to move the mouse handling at this point. Maybe there is a chance to revamp this later. But right now we can just pass everything we need to the method.

I can elaborate more on this at a later point in time if need be, but I'm a bit short on spare time atm.

RobertMueller2:
I have one additional idea:

If you have a look at #385, when merged, this could allow an early exit trigger for the Undo action. Perhaps we could call it "undo-action-when-undo-stack-emtpy" (experimental) for lack of better description. This would achieve an UndoOrExit, but doesn't need an extra UndoOrExit action. If that explanation made sense :)

robertwidfen:
I was wondering about the rationale behind having `undo` as an action. Is it to allow binding it to a specific key or button? 

IMHO it would be better to allow generic bindings - see #450 I hope to make a PR for this soon, but it is only for keys so far not for the mouse.

Also, regarding `undo-or-exit` - I would recommend keeping these as separate actions instead of overloading one - here I also do not understand the rationale behind it.

Issue #89, opened 2024-05-25
[feature] link upload and/or additional buttons with custom functions


evesdropper:
Wondering if it'd be possible to have a button that uploads to an image server (e.g. I currently have a script that uploads to imgbb) and copies the link onto clipboard like how lightshot (which I used on Windows in the past) worked.

This might force people to use one specific image server so I was thinking if not then maybe it could be possible to put additional buttons/keymaps that could map to custom functions such as uploading to an image server.

gabm:
I am not against it. But I am not sure, if we have the time to develop and maintain those features?

sand4rt:
Yeah It would be great if there was an option to add custom keybindings. In swayimg i have a "Mail image" _(opens a mail client with the image attached)_ and "Ripdrag" _(drag‑and‑drop)_ keybindings that i'd like to use in satty.

PR #488, opened 2026-04-14
feat: add actions-on-double-click option (experimental)

### LINKED ISSUE #486: Add actions-on-double-click option (experimental)
## Description

Add `--actions-on-double-click` as a new configurable action list, consistent with the existing `actions-on-enter`, `actions-on-escape`, and `actions-on-right-click` options.

When `actions-on-double-click` is not set (empty), double-click does nothing.

Marked as **experimental** since the future implementation of #89 (generic trigger-action bindings) may provide a more general mechanism for this.

### Use case
Double-click to confirm/save is a common gesture in screenshot annotation workflows. Both [Snipaste](https://www.snipaste.com/) and [PixPin](https://pixpinapp.com/) use double-click to confirm/save in their annotation modes.

### Implementation notes
- Detected via `n_pressed >= 2` on primary button click.
- Routed through `process_actions` (from #485), so all action variants including `undo` and `undo-or-exit` work.
- No fallback behavior when unconfigured — simply does nothing.

Depends on: #485
Related discussion: #481

clansty:
## Summary

- Add `--actions-on-double-click` as a new configurable action list (experimental)
- When not configured, double-click does nothing (no fallback to `actions-on-enter`)
- Marked experimental since #89 may provide a more generic mechanism
- Depends on #487 (undo action) for `process_actions` infrastructure

Closes #486
Related discussion: #481

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,13 +142,13 @@ primary-highlighter = "block"
 # Disable notifications
 disable-notifications = false
 # Actions to trigger on right click (order is important)
-# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit]
+# [possible values: save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit, undo, undo-or-exit]
 actions-on-right-click = []
 # Actions to trigger on Enter key (order is important)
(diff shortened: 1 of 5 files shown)


RobertMueller2:
see #486

PR #494, opened 2026-04-28
feat: make femtovg area transparent

### LINKED ISSUE #402: Change the black application background
It would be nice if there were an option to change the application’s black background color. I tried to do this using `overrides.css`, but couldn’t find a CSS selector.

RobertMueller2:
Closes: #402

This adds transparent background to femtovg area and adds a css class "inner_box" to the GtkBox surrounding it, which is styled with black bg by default, thus allowing to override the color.

To get transparency for all of it, window, .inner_box and .outer_box woudl
 need to be styled as transparent via overrides.css.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ Satty ships with [minimal builtin CSS](https://github.com/Satty-org/Satty/tree/m
 }
 ```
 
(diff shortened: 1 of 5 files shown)


robertwidfen (approved):
LGTM

Thanks.

PR #498, opened 2026-05-06
fix: support TTC font face selection via fontconfig index

### LINKED ISSUE #495: font error
### Version

0.20.1

### Description

I configured noto sans cjk sc as the main font in the config file, and then also put noto sans cjk sc on top of noto sans cjk jp in fallback, but the typing is still the style in noto sans sjk jp instead of noto sans cjk sc

```
# Font to use for text annotations
[font]
family = "Note Sans CJK SC"
style = "Regular"
# specify fallback fonts (0.20.1)
# Please note, there is no default setting for these and the fonts listed below
# are not shipped with Satty but need to be available on the system.
fallback = [
    "Noto Sans CJK SC",
    "Noto Sans CJK JP",
    "Noto Sans CJK TC",
    "Noto Sans CJK KR",
    "Noto Serif CJK SC",
    "Noto Serif JP",
    "IPAGothic",
    "IPAexGothic",
    "Source Han Sans"
]
```

Image

sitiyou:
See #495 for more information.

Satty currently loads fonts through `Canvas::add_font`, which always uses face index `0` internally.

This PR changed the font initialization flow to:
- create a `TextContext` first
- load the matched font through `TextContext::add_shared_font_with_index`
- initialize the canvas with `Canvas::new_with_text_context(...)`

By doing so, Satty can use the face index returned by fontconfig and load the correct TTC face instead of always falling back to face `0`

---

close #495

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -28,7 +28,7 @@ use crate::{
     tools::{CropTool, Drawable, Tool},
 };
 
-use super::set_font_stack;
+use super::{font_stack, set_font_stack};
 
 const TRANSPARENCY_SQUARE_SIZE: usize = 64;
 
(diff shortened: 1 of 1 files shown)


sitiyou:
I have fixed the ci linter errors

RobertMueller2:
Thank you looking into this and for the PR. I first thought this was a femtovg bug, but this change looks almost like we're not using it correctly?

I haven't tested the change fully yet, but I've got two remarks.
- can you use `eprintln!` instead of `println!` please? we're trying to correct this in #479 for all the other places
- the two match arms do almost the same. Perhaps this could be deduplicated with a helper function e.g. with is_fallback bool parameter? What do you think?

sitiyou:
> I first thought this was a femtovg bug, but this change looks almost like we're not using it correctly?

I agree this is not really a femtovg bug.
`Canvas::add_font` as a simple high-level API is reasonable even without a `face_index` parameter, since `TextContext` already provides the full font-loading API.

> I haven't tested the change fully yet, but I've got two remarks.

I have replaced my `println!` with `eprintln!`, and created a `load_font` closure that consolidates all font loading logic.

I also changed `loaded_paths` from `HashSet<PathBuf>` to `HashSet<(PathBuf, u32)>`, using (path, face_index) as the deduplication key so different faces in the same TTC file are not mistaken for the same font.

---

It now works great on my system. I'm happy to help with testing if you need it.

RobertMueller2:
Makes sense and looks good to me.

PR #499, opened 2026-05-09
feat: prefill save-as filename and remember folder

### LINKED ISSUE #462: save-as - default path and filename patern
### Description

On save-as, the file picker would open in a default path and the "name" field would be filled with a default filename pattern (Ex: test-%Y-%m-%d.png). 
This is useful when you need to name the screenshot but still want a timestamp of when you took it (Ex: article/book paragraphs). 

Excellent tool otherwise.

Vladkarok:
Closes #462.

This improves the Save As dialog by using the configured `output-filename` as the initial filename/path when available. This lets timestamp patterns such as `test-%Y-%m-%d.png` prefill the file chooser name while still allowing the user to edit it before saving.

It also remembers the last folder selected via Save As and reuses it as the initial folder next time, falling back to the `output-filename` parent directory when no remembered folder exists.

Verification:
- `cargo fmt --all`
- `cargo test save_as`
- `cargo check`
- `cargo clippy --all-targets --all-features -- -D warnings`

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -405,7 +405,17 @@ impl SketchBoard {
     }
 
     fn save_as_initial_dir(configured_output_path: Option<&Path>) -> Option<PathBuf> {
-        if let Some(last_dir_file) = Self::save_as_last_dir_file()
+        Self::save_as_initial_dir_from_file(
+            Self::save_as_last_dir_file().as_deref(),
+            configured_output_path,
+        )
+    }
+
+    fn save_as_initial_dir_from_file(
+        last_dir_file: Option<&Path>,
+        configured_output_path: Option<&Path>,
+    ) -> Option<PathBuf> {
+        if let Some(last_dir_file) = last_dir_file
             && let Ok(last_dir) = fs::read_to_string(last_dir_file)
         {
             let last_dir = PathBuf::from(last_dir.trim());
@@ -424,6 +434,10 @@ impl SketchBoard {
         let Some(last_dir_file) = Self::save_as_last_dir_file() else {
             return;
         };
+        Self::remember_save_as_dir_to_file(&last_dir_file, output_filename);
+    }
+
+    fn remember_save_as_dir_to_file(last_dir_file: &Path, output_filename: &str) {
         let Some(parent) = Path::new(output_filename).parent() else {
             return;
         };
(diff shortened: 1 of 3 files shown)


RobertMueller2:
Thank you for the PR. This is not a review yet which might take a bit more time.

I'm not sure yet I'm happy with persisting the last used directory beyond the session. Perhaps I'd prefer a save as filename template config/parameter. Although maybe there can be both, as long as the behaviour is intuitive. Please don't do anything just yet, I have to think about this. :smile:

If we do go with persisting this, I don't think `XDG_CONFIG_HOME` is the correct place. IMHO, this is application state rather than config, and thus should be placed in `XDG_STATE_HOME`. https://specifications.freedesktop.org/basedir/latest/

src/sketch_board.rs
```diff
@@ -365,6 +366,89 @@ impl SketchBoard {
         relm4::main_application().quit();
     }
 
+    fn configured_output_path() -> Option<PathBuf> {
```
RobertMueller2, line 401: This function partially duplicates from `handle_save`. Is there any way to reuse this function in `handle_save`?

src/sketch_board.rs
```diff
@@ -365,6 +366,89 @@ impl SketchBoard {
...
+    }
+
+    fn save_as_initial_dir(configured_output_path: Option<&Path>) -> Option<PathBuf> {
+        Self::save_as_initial_dir_from_file(
+            Self::save_as_last_dir_file().as_deref(),
+            configured_output_path,
+        )
+    }
+
+    fn save_as_initial_dir_from_file(
```
RobertMueller2: Maybe I'm missing something, but can `save_as_initial_dir` and `save_as_initial_dir_from_file` be combined?

src/sketch_board.rs
```diff
@@ -365,6 +366,89 @@ impl SketchBoard {
...
+    }
+
+    fn remember_save_as_dir(output_filename: &str) {
+        let Some(last_dir_file) = Self::save_as_last_dir_file() else {
+            return;
+        };
+        Self::remember_save_as_dir_to_file(&last_dir_file, output_filename);
+    }
+
+    fn remember_save_as_dir_to_file(last_dir_file: &Path, output_filename: &str) {
```
RobertMueller2: And can `remember_save_as_dir_to_file` be combined with `remember_save_as_dir`?

src/sketch_board.rs
```diff
@@ -365,6 +366,89 @@ impl SketchBoard {
...
+                    home_dir.push(tilde_stripped);
+                    output_filename = home_dir.to_string_lossy().into_owned();
+                }
+
+                Some(PathBuf::from(output_filename))
+            })
+    }
+
+    fn save_as_last_dir_file() -> Option<PathBuf> {
+        let config_home = std::env::var_os("XDG_CONFIG_HOME")
```
RobertMueller2: As mentioned earlier, this should probably use `XDG_STATE_HOME` instead.

RobertMueller2 (commented):
I have made up my mind. Let's leave the dir remembering function in.

I'll add a followup issue for an explicit pattern later on, which can be implemented later.

Vladkarok:
Thanks, agreed on all points.
  
  I just used in my workflow both fixes so combined them in one PR but separating them will be definitely better.

  I'll update this PR to:
  - reuse the existing output filename resolution path instead of duplicating `handle_save` logic
  - store the remembered Save As directory under `XDG_STATE_HOME` via the existing `xdg` crate
  - simplify/rename the helper functions so they don't look like test-only wrappers leaking into production

  I'll push the fixes to this branch.

Vladkarok:
Thanks, agreed on the review points. I pushed a follow-up commit that reuses the output filename resolution from both save paths, stores the remembered Save As directory under XDG_STATE_HOME via the existing xdg crate, and simplifies/renames the helper functions so they are less test-wrapper-like.\n\nVerification run locally:\n- cargo fmt --all\n- cargo test save_as\n- cargo check\n- cargo clippy --all-targets --all-features -- -D warnings

RobertMueller2:
Ok, thanks. I got three final remarks, I think.

- This change introduces two new `println!` occurrences. We're trying to replace these with `eprintln!` for better consistency in #479, but it would be good if other changes wouldn't introduce new ones in the meantime. Can you please use `eprintln!` in the new sections?
- There's two occurrences of `"save_as_last_dir"` (outside of test), could this be a CONST?
- When we read the state file, we don't _really_ know what's in there. `is_dir()` is a decent plausibility check. What if the file contents is very long, e.g. from corruption? In that case I think we could skip `read_to_string`, e.g. if it's longer than an arbitrary long value of 10000 bytes which should be a reasonable unreachable upper limit for all systems. What do you think?

RobertMueller2:
I suppose for the last point one might argue that we don't have such checks elsewhere ;)

Vladkarok:
> Ok, thanks. I got three final remarks, I think.
> 
> * This change introduces two new `println!` occurrences. We're trying to replace these with `eprintln!` for better consistency in [fix: replace println! with eprintln! #479](https://github.com/Satty-org/Satty/pull/479), but it would be good if other changes wouldn't introduce new ones in the meantime. Can you please use `eprintln!` in the new sections?
> * There's two occurrences of `"save_as_last_dir"` (outside of test), could this be a CONST?
> * When we read the state file, we don't _really_ know what's in there. `is_dir()` is a decent plausibility check. What if the file contents is very long, e.g. from corruption? In that case I think we could skip `read_to_string`, e.g. if it's longer than an arbitrary long value of 10000 bytes which should be a reasonable unreachable upper limit for all systems. What do you think?

Thanks, addressed these in the latest push.

Changes:
  - switched the two newly added diagnostics from `println!` to `eprintln!`
  - introduced constants for the **Save As** state filename and max state-file size
  - added a metadata guard before reading the remembered directory state file, skipping it if it is not a regular file or is larger than 10 000 bytes

Vladkarok:
> I suppose for the last point one might argue that we don't have such checks elsewhere ;)

Agreed, it is defensive, but cheap and reasonable for this tiny state file :)

RobertMueller2:
lgtm, thanks :)

PR #508, opened 2026-05-21
fix: Ctrl+Y for DE by removing layout-independent keys

robertwidfen:
Remove layout-independent key mappings that caused "Ctrl+Y" to malfunction in the DE locale. On German keyboards, the 'Z' and 'Y' keys are swapped. Because of this, 'Y' was incorrectly matching KeyMappingId::UsZ, triggering an "Undo" instead of a "Redo".

Global layout adjustments like this should be handled at the system or framework level rather than per application.

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -1077,7 +1077,9 @@ impl Component for SketchBoard {
                         ToolUpdateResult::StopPropagation
                         | ToolUpdateResult::RedrawAndStopPropagation => active_tool_result,
                         _ => {
-                            if ke.is_one_of(Key::z, KeyMappingId::UsZ)
+                            if ke.key == Key::y && ke.modifier == ModifierType::CONTROL_MASK {
+                                self.handle_redo()
+                            } else if ke.is_one_of(Key::z, KeyMappingId::UsZ)
                                 && ke.modifier == ModifierType::CONTROL_MASK
                             {
                                 self.handle_undo()


RobertMueller2:
Funny I should have missed this, I'm a German using a US keyboard and switching input language all the time. You're definitely onto something there. But before we can get this merged, I need to find out why and under which circumstances it was implemented like that.

robertwidfen:
Yeah, I was puzzled why it was not working and when I looked at the code I was puzzled even more. 

I did not dig into history to find out why it was implemented and first was thinking to just switch the order of the checks for C-Z and C-Y which would fix it for this case. But this eventually may cause similar issues on other layouts, so I think it is best to remove it. IMHO if I would want to always have US-layout bindings on other layouts it should be handled not in the app, but on a level where all apps would "benefit" from it.

robertwidfen:
This regression was introduced with PR #147 fixing the issue #121 (shortcuts to be available on non-latin keyboards).  It is a general issue not unique to satty, e.g. see https://gitlab.gnome.org/GNOME/gtk/-/work_items/5384 ... and many layouts suffer from this, e.g. in French layout Q and A are switched and so C-A eventually triggers quit when bindings are not handled correctly ...

Likely using https://docs.gtk.org/gtk4/class.ShortcutController.html fixes the original issue by falling back to physical US layout if no match was found, but I could not find a reference to prove this. 

Also rewriting this PR to do a two step check for all bindings is IMHO to much - so I changed it to just test C-Y additionally before C-Z.

RobertMueller2:
Thanks for the research. I hope to have a closer look towards the weekend :)

RobertMueller2:
ok, I think I understood now. LGTM, thanks :)

robertwidfen:
Thanks for merging - Uhh, to fast you were resp. I have been to slow ... 🙃 

I have updated the description as it was not matching the code anymore.
But this was pushed just now after your merge.

See https://github.com/robertwidfen/Satty/commit/c9c0c0f5e3c7c8b8f6cb986e72ccb1a7220336e3

How to solve this? 🤔

RobertMueller2:
Oops, sorry. Just found some time between two customers, probably won't happen again this week. :laughing: 

I think we can live with that. If anyone even reads it and ends up being confused, they can have a look at the PR.

PR #509, opened 2026-05-21
feat: select custom color for invalid palette idx

robertwidfen:
If index in palette is out of range select custom color.


RobertMueller2:
How do I test this, please? I thought commenting out palette in config and clearing gtk custom-color setting was sufficient, but it doesn't seem to do the trick.

robertwidfen:
It should not be necessary to change config in any way. Hmm odd.

I have 5 default colors which I can switch by 1-5 key - any key other numeric kay will select the custom color. If you palette is 9 colors then only 0 will switch to custom color.

RobertMueller2:
Thanks for elaborating. Yes, I understand it now and it works well.

LGTM, thanks! :)

PR #475, opened 2026-04-02
fix: input-scale ignored with fullscreen and initial resize

RobertMueller2:
Closes: #420

PR #510, opened 2026-05-21
feat: add initial-zoom config option

robertwidfen:
Allows starting with 1:1 zoom.


RobertMueller2:
I'm fairly sure #475 addresses this already. It hasn't been merged yet.

robertwidfen:
You are right - I tested your fix, works as expected - scale vs. zoom 🙃 should have seen it there is an config option for this already.

PR #512, opened 2026-05-23
feat: adjust annotation size by scrolling over the toolbar button

### LINKED ISSUE #171: Feature request / Scroll on annotation size to change value

### LINKED ISSUE #513: feature: Allow adjusting annotation size by scrolling over the toolbar button
### Description

Currently, the annotation size factor is adjusted by clicking the toolbar button and editing the value in the dialog.

It would be useful if hovering the annotation size button and using the mouse wheel adjusted the value directly in small steps, while keeping the dialog for exact input.

This would make quick size adjustments easier without changing the existing dialog-based workflow.

sonnymilton:
## Summary

  This changes the annotation size button in the bottom toolbar so the scroll can adjust the annotation size directly while the cursor is hovering the button.

  The change keeps the existing annotation size dialog intact and reuses the same `ToolbarEvent::AnnotationSizeChanged` flow for updates.

  ## Notes

  - No new dependencies were added.
  - No README changes were needed because this does not change documented CLI or configuration behavior.
  - closes #513 
  - closes #171

### DIFF
--- a/src/ui/toolbars.rs
+++ b/src/ui/toolbars.rs
@@ -72,6 +72,7 @@ pub enum StyleToolbarInput {
     SetVisibility(bool),
     ToggleVisibility,
     ShowAnnotationDialog,
+    AdjustAnnotationSize(f32),
     AnnotationDialogFinished(Option<f32>),
     DimensionsChanged((i32, i32)),
 }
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Thanks for the PR. We'll probably merge this after we had a chance to review. But please note that this might be temporary, there's a (currently inactive) PR #178 which may revamp the S/M/L and annotation size with a different approach and control.

RobertMueller2:
>     * If there is a related issue, please link it here to follow the contribution guidelines.

:laughing:

sonnymilton:
Thanks for the note. That makes sense.

  I’m fine with this being treated as a temporary improvement if PR #178 ends up replacing this control with a different approach.

Created a related issue: #513

RobertMueller2:
LGTM, thanks :)

PR #518, opened 2026-05-26
fix: do not create empty text objects

robertwidfen:
Also removes code duplication by adding get_text().

Clicking with the text tool will already generate a drawable also if the text is empty. It is kind of useless, but does no real harm. But with the editing feature as discussed in #159 this leads to odd small objects that can be accidentally selected.


RobertMueller2:
LGTM, Thanks! :)

PR #521, opened 2026-05-30
fix: broken icon in scale dialog

robertwidfen:
Was somewere lost ... or maybe not provided by my Arch system/setup with Hyprland.

image

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -73,6 +73,7 @@ fn main() -> Result<(), io::Error> {
             "paint-bucket-regular",
             "page-fit-regular",
             "resize-large-regular",
+            "arrow-counterclockwise-regular",
         ],
     );
 
--- a/src/ui/toolbars.rs
+++ b/src/ui/toolbars.rs
@@ -842,7 +842,7 @@ impl Component for AnnotationSizeDialog {
                     set_hexpand: false,
 
                     set_tooltip: "Reset Annotation Size Factor",
-                    set_icon_name: "edit-reset-symbolic",
+                    set_icon_name: "arrow-counterclockwise-regular",
                     connect_clicked[sender] => move |_| {
                         sender.input(AnnotationSizeDialogInput::Reset);
                     },


RobertMueller2:
hmm, it works for me. But there *is* something wrong with it, it's not included in build.rs, which explains why it might not be available everywhere, so let's try to fix that.

reset-icon

That said, I'm kinda inclined not to use the same icon we use for undo in the toolbar, because here it's not really an undo but reverting the setting to what was configured. But we should use something that can be included via build.rs. There is `revert`, but that doesn't look right to me. How about `undo-fill` maybe? That looks kinda close to what we had.

robertwidfen:
Yeah "undo" it is not perfect, actually I was looking for a anti-clockwise arrow, but could not find it.

undo-fill-symbolic.svg looks different for me - odd

image

What is a good way to browse the available icons? I just now realized it is much easier not to rely on some web site, directly browse ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/relm4-icons-build-0.11.0/icons/ 

How about `arrow-counterclockwise-regular` - I actually do not know how the original looked like 😉

image

RobertMueller2:
> Yeah "undo" it is not perfect, actually I was looking for a anti-clockwise arrow, but could not find it.
> 
> undo-fill-symbolic.svg looks different for me - odd
> 
> image
> 
> What is a good way to browse the available icons? I just now realized it is much easier not to rely on some web site, directly browse ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/relm4-icons-build-0.11.0/icons/ 

Not near a computer so can't dig out a link, but I'm using the md in the crate and then browse a clone.

It's not perfect, and then down the road I'm sure somebody is going to point out they're not all from the same category and therefore don't match. But... I think somebody else will have to investigate that. 😄

> How about `arrow-counterclockwise-regular` - I actually do not know how the original looked like 😉

Works for me. Until we introduce rotate 😂

PR #522, opened 2026-05-30
fix: manpage needs target directory in Makefile

RobertMueller2:
Without the `-t` the last element is not treated as directory. And if `${PREFIX}/share/man/man1` doesn't already exist, `satty.1` is copied to a file named `man1`. This was a copy+paste mistake, the PR fixes this by adding the missing  `-t` argument.


robertwidfen (approved):
👍🏻

Issue #395, opened 2026-01-20
Fix stuff for crates.io publishing


RobertMueller2:
- [x] Adjust README to point out cargo install/binstall
- [ ] Crate ownership
- [x] address `cargo publish` errors, warnings etc
- [ ] add github action to publish

PR #526, opened 2026-06-04
fix: release.nu to also replace satty_cli dependency

RobertMueller2:
with working towards #395 (cargo publish), we needed to explicitly add satty_cli's version to main Cargo.toml or cargo publish throws a warning. But we forgot that now this needs to be replaced by release.nu.


robertwidfen (commented):
LGTM

PR #530, opened 2026-06-05
chore: updating deps in release.nu becomes optional

### LINKED ISSUE #529: release.nu - optional update of 3rd party deps
Right now, `release.nu` updates `Cargo.lock` with `cargo generate-lockfile`. 

When dealing with #427 bugfix release, this had the unwanted side effect of reverting a revert.

Change `release.nu` so it always updates `satty` and `satty_cli` in `Cargo.lock`. Updating of other packages in there should be optional (input with default "N") -- nowadays we use dependabot to udpate dependencies and don't really need this within `release.nu` any longer.

RobertMueller2:
Closes: #529

see #529 for rationale. 

The variable is needed because `or` in nushell is evaluated eagerly.

### DIFF
--- a/release.nu
+++ b/release.nu
@@ -21,7 +21,13 @@ export def main [version: string] {
     patch_cargo_toml $requested_version
 
     # update cargo lock
-    update_cargo_lock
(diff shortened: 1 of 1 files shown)


robertwidfen (approved):
LGTM

PR #532, opened 2026-06-06
fix: pan with middle mouse button in zoom != 1.0

robertwidfen:
When dragging with a zoom != 1.0 the pixel where dragging started does not stay under the mouse pointer as the event.pos used for the pan is in image coordinates.

Thus add a new event.screen_pos to the MouseEventMsg and keep event.pos which is transformed to image coordinates for tool in handle_event_mouse_input().


RobertMueller2:
interesting, nice find. LGTM.

Issue #338, opened 2025-11-13
Feature: zooming shortcuts/usability


RobertMueller2:
Followup for #336  
 
* add keyboard shortcuts for zooming, perhaps for a few preselected scale factors as well as +/-
* touchpad support

unikitty37:
I use a trackpad, which does not make it easy to click the middle mouse button. Could the keyboard shortcut for panning be holding down the Space and dragging? (This is a de-facto standard, since this is what Photoshop does).

RobertMueller2:
> I use a trackpad, which does not make it easy to click the middle mouse button. Could the keyboard shortcut for panning be holding down the Space and dragging? (This is a de-facto standard, since this is what Photoshop does).

I agree it should be more accessible under these circumstances. While Alt+Cursor works, it's not the same as dragging.

That said, my first thought is to use something like Alt because that's actually a modifier according to GTK (space isn't) and should be easier to evaluate. Just a hunch, haven't actually tried it.

If whoever ends up implementing this finds that using Space as a modifier here is totally easy to do, then fine with me.

robertwidfen:
https://github.com/robertwidfen/Satty/commits/feat_touchpad_zoom_and__pan/

Zooming with mouse now also requires <kbd>Ctrl</kbd> like in other apps.
If image fits then panning is disabled.

There is currently a hard-coded reduction of pan_step_size by -8, not sure if this must be configurable. 🤔 

Mouse:
- <kbd>Ctrl+Wheel</kbd>: Zoom in/out
- <kbd>Middle</kbd>: Pan

Touchpad (2 finger scroll):
- <kbd>Ctrl+Scroll</kbd>: Zoom in/out
- <kbd>Scroll</kbd>: Pan

robertwidfen:
#555 has Ctrl+plus/minus for zooming
#606 has Ctrl-f (fit to window), Ctrl-1 (1:1 original scale), Ctrl-2, ...

Can we close this issue?

PR #535, opened 2026-06-10
feat!: touchpad scroll for zoom (with Ctrl) and pan

robertwidfen:
See issue #338.

Breaking change: Zooming with mouse now also requires Ctrl like in many other apps.

### DIFF
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -35,6 +35,9 @@ const SAVE_AS_LAST_DIR_MAX_BYTES: u64 = 10_000;
 #[derive(Debug, Clone)]
 pub enum SketchBoardInput {
     InputEvent(InputEvent),
+    PinchStart,
+    PinchScale(f32),
+    PinchEnd,
     ToolbarEvent(ToolbarEvent),
     RenderResult(RenderedImage, Vec<Action>),
     RenderResultFollowup(Option<Pixbuf>, Vec<Action>, Option<String>),
@@ -106,6 +109,7 @@ pub struct MouseEventMsg {
     pub button: MouseButton,
     pub modifier: ModifierType,
     pub screen_pos: Vec2D,
+    pub is_touchpad: bool,
     pub pos: Vec2D,
     pub n_pressed: i32,
     pub release: bool,
@@ -127,6 +131,7 @@ impl SketchBoardInput {
             modifier,
             screen_pos: pos,
             pos,
+            is_touchpad: false,
             release,
         }))
     }
@@ -146,14 +151,20 @@ impl SketchBoardInput {
         SketchBoardInput::CommitEvent(event)
     }
 
-    pub fn new_scroll_event(delta_y: f64) -> SketchBoardInput {
+    pub fn new_scroll_event(
+        delta_x: f64,
+        delta_y: f64,
+        modifier: ModifierType,
+        is_touchpad: bool,
+    ) -> SketchBoardInput {
         SketchBoardInput::InputEvent(InputEvent::Mouse(MouseEventMsg {
             type_: MouseEventType::Scroll,
             button: MouseButton::Middle,
             n_pressed: 0,
-            modifier: ModifierType::empty(),
-            screen_pos: Vec2D::new(0.0, delta_y as f32),
-            pos: Vec2D::new(0.0, delta_y as f32),
+            modifier,
+            screen_pos: Vec2D::new(delta_x as f32, delta_y as f32),
+            pos: Vec2D::new(delta_x as f32, delta_y as f32),
+            is_touchpad,
             release: false,
         }))
     }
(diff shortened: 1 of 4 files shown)


src/sketch_board.rs
```diff
@@ -223,11 +231,21 @@ impl InputEvent {
...
+                    if me.modifier.contains(ModifierType::CONTROL_MASK) {
+                        let factor = APP_CONFIG.read().zoom_factor();
+                        match me.pos.y {
+                            v if v < 0.0 => renderer.set_zoom_scale(factor),
+                            v if v > 0.0 => renderer.set_zoom_scale(1f32 / factor),
+                            _ => {}
+                        }
+                    } else {
+                        let pan_step_size = if me.is_touchpad {
+                            -2.0 // invert for natural scrolling and decrease a bit
```
robertwidfen, line 249: Inverting seem natural to me, but I am not sure if the factor 2 only makes it feel right for me on my hardware - testing by others would be great.

RobertMueller2:
Ok, I definitely need more time to test this, might take a few days.

When using the touchpad, it does not feel the same like e.g. in a browser. I need to educate myself about touchpad gestures, though. But my gut feeling is that zooming should be a pinch move without ctrl? Can we figure that out from GTK?

Regarding regular mouse:
- At first glance, I'm not sure I like the viewport clamping (other than ctrl modifier, which definitely makes sense). Before this PR, I can just move the view around freely, with this PR it feels quite limiting. EDIT: I assume the viewport clamping is a bit of a necessity for touchpad usage. I think we could make this configurable somehow, config sets which input types use viewport clamping and default would be to do it just for touchpad? Does that make sense?
- Secondly, even if I'm zoomed in and CAN pan both vertically and horizontally, it feels sluggish and the motion frequently gets interrupted. I think I might accidentally trigger a wheel scroll event which interrupts the panning, need to confirm this.

robertwidfen:
No hurry at all - take your time with the testing, and thank you for the feedback!

The clamping was intended to prevent the screenshot from moving completely "off-screen", which didn't feel too restrictive during my testing. Could you share a bit more about your workflow and why you'd prefer to move further? I'd love to avoid adding another configuration option if possible, as we already have quite a few. 🙃  Maybe I just remove the clamping code 😉, because less code is better and one can find the way back with the resize-to-regular or fit-to-page toolbar buttons.

It's also interesting that it feels sluggish for you - I haven't noticed any lag myself - see video below.

As for pinch-to-zoom: Chromium and other apps do support both pinch and Ctrl+Scroll. I rarely use pinch gestures on my laptop (mostly just on my phone), so I completely missed that, but likely people expect it.

https://github.com/user-attachments/assets/2b04d63d-40ba-48ae-9153-55253e9281ae

src/sketch_board.rs
```diff
@@ -223,11 +231,24 @@ impl InputEvent {
...
                 MouseEventType::Scroll => {
-                    let factor = APP_CONFIG.read().zoom_factor();
-                    match me.pos.y {
-                        v if v < 0.0 => renderer.set_zoom_scale(factor),
-                        v if v > 0.0 => renderer.set_zoom_scale(1f32 / factor),
-                        _ => {}
+                    if me.modifier.contains(ModifierType::CONTROL_MASK) {
+                        let mut factor = APP_CONFIG.read().zoom_factor();
+                        if me.is_touchpad {
+                            factor = 1.0 + (factor - 1.0) / 2.0; // decrease a bit
```
robertwidfen, line 240: See below

robertwidfen:
It is now without clamping.

RobertMueller2:
> No hurry at all - take your time with the testing, and thank you for the feedback!
> 
> The clamping was intended to prevent the screenshot from moving completely "off-screen", which didn't feel too restrictive during my testing. Could you share a bit more about your workflow and why you'd prefer to move further? I'd love to avoid adding another configuration option if possible, as we already have quite a few. 🙃 Maybe I just remove the clamping code 😉, because less code is better and one can find the way back with the resize-to-regular or fit-to-page toolbar buttons.

Well, maybe there wasn't enough time in the morning :)

I tried:
- scroll vertical with mousewheel - feels natural
- scroll horizontal with mousewheel - feels natural
- scroll vertical with touchpad - feels natural
- scroll horizontal with touchpad - feels natural

And I think these four actions actually did benefit from clamping. Sorry :)

However, and this is what I complained about this morning, when I'm using the middle mouse button to pan (and I might  continue to do that) the clamping feels very limiting.

I think my preference would be to use clamping for panning via scrolling, and no clamping for panning via middle mouse button. Does that make sense?

- scroll vertical with touchpad + control to zoom - feels usable, I'm not sure about the expected speed here

> It's also interesting that it feels sluggish for you - I haven't noticed any lag myself - see video below.

The problem isn't lag. I think what happened (but I haven't confirmed via debugger) is this: I started to pan via middle mouse button and tried to move around, then accidentally triggered a scroll wheel event and the panning via middle mouse button stopped or at least got interrupted briefly. This is not what happens before the PR, although I'm sure I hit the scroll wheel just the same, the wheel is quite sensitive.

> As for pinch-to-zoom: Chromium and other apps do support both pinch and Ctrl+Scroll. I rarely use pinch gestures on my laptop (mostly just on my phone), so I completely missed that, but likely people expect it.

Yeah, I think that could be a good addition. :)

robertwidfen:
Now with pinch-to-zoom. 🤏

I dare to resurrect clamping - as discussed resize-to-regular or fit-to-page will bring you back if you got lost and it was irritating for you and likely is for others if you do not know the reason for the invisible bounds.

src/sketch_board.rs
RobertMueller2: The value works for well for me, but I understand the concern about different hardware.

I know we have many options already, but I think this could benefit from one. `pan_touchpad_step_size`?

RobertMueller2:
Fair enough :) looks good to me now and pretty much ready to merge, should be fine once the inline comment is resolved.

RobertMueller2:
I'm going to merge this now. If anyone is bothered by the hardcoded value and inversion and/or the zoom factor modification, they can of course open an issue for these, and then we can talk about whether there's a better default, or this needs option(s).

src/sketch_board.rs
robertwidfen: I would also make one for zoom.

Command line options also?

robertwidfen:
If we might need config in the future it is at [robertwidfen:feat/touchpad_zoom_pan_config](https://github.com/robertwidfen/Satty/tree/feat/touchpad_zoom_pan_config) 

I will keep it on my stack for a while, but hope that we do not need it 😉

Issue #243, opened 2025-06-17
Numbered marker seems too big compared to others tool (line for example)


MangelMaxime:
Hello,

It feels like the numbered marker much bigger than the other tool and compared to text on Screenshot.

Trying to adjust the Size factor make it possible to reduce its size, however the others tool seems too small after:

![Image](https://github.com/user-attachments/assets/7075eb31-052c-4004-a890-7f35fef5b97b)

Is there a way to set the size of the numbered marker (or each tools) independently? Is it possible to revisit its default size?

RobertMueller2:
I think #178 changes the defaults for the tools, feel free to try that out and add feedback to the existing PR.

Adjusting tools individually is currently not possible and not planned. Exposing these values via config doesn't seem terribly difficult once #178 is done and merged, but somebody would have to implement that. Adjusting the same via GUI would be a logical next step after, but personally I think before a GUI revamp that would be a wasted effort.

MangelMaxime:
I tried to run `cargo run` from the PR you linked but it fails to load Satty correctly so I can't test it. I am not expert in Rust so perhaps this is not how I should run it.

![Image](https://github.com/user-attachments/assets/55131bf6-cdeb-406e-a43f-c2170ec394e6)

But I don't think this PR will solve the problem, because if I understood, it changes how S, L, M sizing works but not the fact that Numbered Marker are way too big compared to others tools. When you look at the default size of the marker 1 it is not really usable.

robertwidfen:
https://github.com/robertwidfen/Satty/tree/refactor_smaller_marker

The additional ring is also to much fuss for me - covering too much.
Of course I am biased coming from other screenshot tools.

<img width="472" height="328" alt="Image" src="https://github.com/user-attachments/assets/72bc5524-bf34-4cb4-ae9f-ea312fad27a2" />

RobertMueller2:
> https://github.com/robertwidfen/Satty/tree/refactor_smaller_marker
> 
> The additional ring is also to much fuss for me - covering too much. Of course I am biased coming from other screenshot tools.

Definitely yes from me on the size reduction.

Regarding the outer ring, that might be a matter of preference. Personally, I'm undecided which I like better. How about an alt mode or config? The additional complexity shouldn't be that bad, I think.

robertwidfen:
Good idea with Alt enable the outer ring 🤩👍🏻 - could be used to make the start/end marker to stand out a bit.  I dare adding another config and prefer keep complexity lower.

[...]

PR #539, opened 2026-06-12
refactor: optional extra ring around marker

robertwidfen:
See issue #243.

Make markers smaller and draw additional ring only if Alt is held. Marker is commited at button release and thus stays moveable.


RobertMueller2:
Code looks good, function looks good.

Just two minor things I noticed:
- when dragging the marker and pressing any key which is not bound, ring is toggled once
- when dragging the marker and pressing any key which is a tool shortcut, there's no redraw and the marker remains until next redraw (which might not happen until next tool action). This isn't exclusive to marker though, and perhaps we need to fix it outside of this PR. Not sure if it broke recently, or nobody ever looked close enough :)

robertwidfen:
Thanks for testing 🤗 - indeed the code was odd, I checked for the Alt-modifier where I should check for the keys Alt_L and Alt_R - at a second glance I added a "!" where it did not make sense.

~The tool key issue is actually magically fixed in https://github.com/robertwidfen/Satty/tree/feat_generic_key_bindings #450 - first I was not able to reproduce it as I often work on a merge or all branches and the `jj absorb` fixes. 🙃 But I want to test this a bit more before creating a pull request.~

It is still there, had `StopPropagation` for some minutes in the code, but this also stops tool switching - created #540 for discussion of expected behavior.

RobertMueller2:
lgtm.

PR #542, opened 2026-06-15
refactor: cleanup & convert .vscode configs to templates

robertwidfen:
Cleans up tracked .vscode files and replaces them with .template versions to prevent developers from accidentally committing local debug settings.


RobertMueller2:
lgtm, thanks :)

Issue #118, opened 2024-09-04
A layer-shell mode?


stacyharper:
Hey there,

Satty is really great, thanks for this. But I'm still looking for a similar program that would run as a layer-shell. The idea would be to draw above the outputs, for a presentation by example. I think it would need an additional button to toggle the event listener, so that we could interact with the underlying surfaces.

Is it something someone already thought about? It is doable in the scope of Satty? Or does someone knows about an existing program that would behave this way?

Thanks!

gabm:
I dont know whether its doable.. but why wouldn't fullscreen be enough?

raiguard:
Fullscreen doesn't work well on Sway if you already have another fullscreen window open, i.e. a fullscreen video game. Using layer-shell would fix this issue.

stacyharper:
I meant that instead of giving a filename as argument, you would run Satty above the existing outputs. Then the clipboard button would take a screenshot, apply the user drawing, and yank the resulting image.

jficz:
Exactly what I came here to ask for. Basically make Satty also a _screen_ annotation tool, not just _screenshot_. I can't imagine how hard would that be but I guess just making it layer-shell and transparent should be enough?

gabm:
I am open for PRs, but my time is limited :/ see #138

jficz:
Time to finally learn Rust I guess :)

raiguard:
> I meant that instead of giving a filename as argument, you would run Satty above the existing outputs. Then the clipboard button would take a screenshot, apply the user drawing, and yank the resulting image.

This would also be cool! I think there are usecases for both - my usecase is to be able to annotate screenshots taken when an app is fullscreen on Sway (I currently have to un-fullscreen the game and the fullscreen satty to use it). But live annotation would also be really neat.

jficz:
> If we do live annotation, we somehow have to acquire a screen copy at some point or we can't save the result.

Not necessarily. At least not for the initial functionality.

If just the annotation part is implemented (by running satty without any inputs on the top layer of the screen with transparent background for example), then user should be able to take a screenshot by other means if one is needed.

I understand that this will render the clipboard-copy and `-o` options useless however.

On the other hand, in my opinion, that wouldn't change the scope of Satty, as it is primarily an _annotation_ tool and the file/clipboard management is essentially just a way to _use_ the result of the annotation. On-screen annotation can be thought of as a _use_ on its own which does not necessarily require saving.

gabm:
[...]

PR #546, opened 2026-06-16
feat: span fullscreen="all" across all monitors on Wayland (layer-shell)

FedotCompot:
## Disclaimer

**The whole PR was vibe-coded with Opus 4.8**

## Summary

`fullscreen = "all"` previously only ever covered a single monitor on Wayland. The implementation relied on GDK's `FullscreenMode::AllMonitors`, which is a no-op on the Wayland backend — and a single Wayland surface cannot span multiple outputs by design.

This PR makes `fullscreen = "all"` actually span every monitor on Wayland by using the `wlr-layer-shell` protocol: the app's root window becomes the primary layer-shell surface, and one additional layer-shell surface is created per remaining output. Each surface renders its **own native-scale slice** of the screenshot, so together they cover the whole layout 1:1.

Behaviour is unchanged on X11 (still uses GDK's native all-monitor fullscreen) and it falls back gracefully to a single screen on compositors without layer-shell and for single-monitor setups.

Related to #118 — this introduces `wlr-layer-shell` support to the codebase, a prerequisite for the layer-shell / live-annotation ideas discussed there.

## How it works

- A new per-area "layout view" transform pins each `FemtoVGArea` to a fixed image slice at native scale (`image_origin` + image-pixels-per-device-pixel) instead of the usual fit/center.
- The App root window is configured as the primary layer-shell overlay (anchored to all edges of its monitor); `SketchBoard` creates one more layer-shell window per other monitor.
- Each monitor's area renders the shared scene on its own GL context. Committed drawables are broadcast (cloned) to every area; undo/redo/reset/refresh are broadcast too. This keeps per-canvas state independent (e.g. the blur tool's cached image), so each output renders correctly.
- Secondary monitor areas convert pointer events to image coordinates locally and forward them as image-space events, so drawing works on every monitor.
- Save/copy continues to render the full image on the primary area, so the exported result contains annotations from all monitors.
- The blur tool screenshots the canvas, so it now clamps the sampled region to the framebuffer to avoid out-of-bounds sampling on surfaces that only show a slice.

## New dependency

- `gtk4-layer-shell` (Rust crate `0.7`, targeting gtk4-rs 0.10). Requires the `gtk4-layer-shell` C library at build and runtime.

## Testing

Verified on Hyprland with a 3-monitor layout (1920×1080 @ -1920,0 · 2560×1440 @ 0,0 · 2560×1440 @ 2560,0):

- All three monitors covered, each showing its correct, seamlessly-aligned slice; toolbars overlaid on the primary monitor.
- Drawing verified on every monitor (primary and both secondaries), correctly positioned.
- Save/copy produces the full 7040×1440 image containing annotations drawn on all three monitors.
- Blur verified on a secondary monitor with real content — renders correctly, no crash.
- Single-window mode unchanged (normal toplevel, no layer-shell); `cargo test` passes, `cargo clippy` clean.

## Known limitations (initial version)

- Keyboard / text-tool input is routed to the primary monitor.
- A blur dragged across a monitor seam may look slightly off.
- Mixed-DPI monitors are not fully handled yet (uniform scale — the common case — is correct).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,9 +99,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.12.1"
+version = "2.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "84d7ced0ae9557296835c32bf1b1e02b44c746701f898460fb000d7eaa84f00a"
+checksum = "b4388bee8683e3d04af747c73422af53102d2bd24d9eadb6cbc100baef4b43f8"
 
 [[package]]
 name = "bumpalo"
@@ -141,7 +141,7 @@ version = "0.21.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b01fe135c0bd16afe262b6dea349bd5ea30e6de50708cec639aae7c5c14cc7e4"
 dependencies = [
- "bitflags 2.12.1",
+ "bitflags 2.13.0",
  "cairo-sys-rs",
  "glib",
  "libc",
@@ -160,9 +160,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.2.63"
+version = "1.2.64"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "556e016178bb5662a08681bbe0f00f8e17631781a4dfc8c45e466e4b185ec27f"
+checksum = "dad887fd958be91b5098c0248def011f4523ab786cd411be668777e55063501f"
 dependencies = [
  "find-msvc-tools",
  "shlex",
@@ -333,7 +333,7 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b96028ce3ff03972312fd8243281858e80fc0f9838b1f035676b6c199214d9e"
 dependencies = [
- "gl_generator",
+ "gl_generator 0.9.0",
  "libc",
  "pkg-config",
  "shared_library",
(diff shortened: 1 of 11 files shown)


RobertMueller2:
Thank you for this PR.

I can't speak for all of my maintainer colleagues of course, but personally I don't think that we are going to merge this into main.

My gut feeling is that Satty is not the correct place to solve this problem. This should be solved in Wayland, and it has been discussed in the past already: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/work_items/99  

Yes, that approach is (obviously) very veeeeeeerrrry sllllllllllloooooooooow, but it's the right one. My suggestion for anyone looking for a solution for this is to direct any effort towards the Wayland item mentioned above.

If it was a tiny change providing a workaround, I probably would not mind adding it to Satty. But it is not a tiny change. It would add considerable complexity, and brings along more changes required to even make it work. Take the need for a custom colour dialog, for example. Complexity we'd have to support once we merge. I prefer living with a known issue here.

That said, I think we might be looking into cherry-picking b6a1d3a8a970b8140d150e6a5844f60a68ba54f7, because that's actually a simplification and looks quite useful.

FedotCompot:
Ok, perfectly understandable, if you want I can make a PR with the tool size UI.

Regarding workaround options...

The first thing I tried was a single window spanning all monitors (7040x1440@-1920x0 in my case). But past a certain point the window rendering itself becomes inconsistent, colors shift so IDK. Maybe there is a way of making it work I'm not aware of, I'm really not an expert of rust or desktop applications in general, just a poor guy missing flameshot level ergonomics for screenshots after moving to hypland ahahahah

So I will keep my fork and continue doing some improvements from time to time based on my needs and time available and will use a local build on my machines. Really like the project, hope it will survive the maintainers crisis

RobertMueller2:
> Ok, perfectly understandable, if you want I can make a PR with the tool size UI.

Thank you for understanding.

Yes, if you don't mind, a PR would be nice.

Just a small heads-up: we may revamp this again, so it may just be temporary. The S-M-L plus annotation size is possibly not the greatest UX, but we haven't really decided on anything yet, so I think we'd merge the commit for the time being.

FedotCompot:
Yeah... I also noticed that, me personally I REALLY loved the scrolling with a circle around the cursor as a preview to change the tool size without a specific numer on flameshot, really simple and intuitive.

I'm on vacation 20-28 as soon as I have time when I get back I will try to implement (vibe code) it on my fork, will try to do it on both my layer-shell and master so in case you like it you can consider merging.

Also regarding the color menu it could be an in place popover instead of a separate window. Just a thought...

PR #547, opened 2026-06-16
feat: smaller and same marker size for 1-99

robertwidfen:
Further improve issue #243 and PR #539.

- use same size for 1-99 - also avoids site jitter due to  small differences in text metrics.
- simpler overall size calculation.

image


RobertMueller2:
lgtm.

RobertMueller2 (approved):

robertwidfen:
I just reviewed it my self again and fixed the comment on the jitter - it was a bit messed up and now is clearer.

So it is ready for merge now.

PR #548, opened 2026-06-18
feat: inline annotation-size -/+/reset buttons in the toolbar

FedotCompot:
Replace the separate annotation-size dialog with controls directly in the style toolbar: a minus and plus button (±0.1), a live value readout, and a reset button that restores the configured default. Scrolling over the value still fine-tunes by ±0.01; the scroll controller now uses DISCRETE so one wheel notch is exactly one step instead of jumping erratically with smooth scrolling.

The value still propagates via ToolbarEvent::AnnotationSizeChanged, so nothing downstream changes. Removes the AnnotationSizeDialog component, its input/output enums, the controller field, the show_annotation_dialog method, and the ShowAnnotationDialog/AnnotationDialogFinished messages.

### DIFF
--- a/src/ui/toolbars.rs
+++ b/src/ui/toolbars.rs
@@ -32,12 +32,7 @@ pub struct StyleToolbar {
     visible: bool,
     annotation_size: f32,
     annotation_size_formatted: String,
-    annotation_dialog_controller: Option<Controller<AnnotationSizeDialog>>,
     output_dimensions: String,
-}
-
-pub struct AnnotationSizeDialog {
-    annotation_size: f32,
 }
 
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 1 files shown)


RobertMueller2 (approved):
lgtm.

robertwidfen:
I took your idea - it is a good one, but I use the SpinButton from the dialog and this also allows direct editing. Unfortunately I had to touch a bit more code to handle focus and ...
 
https://github.com/user-attachments/assets/da4f2629-a849-4e97-94da-164d660c3f48

robertwidfen:
For the direct comparison ;-)

image

FedotCompot:
At this point remove the S M L buttons and just use the number as a scale factor directly, no?

robertwidfen:
I am kind of uncertain - I like the S-M-L for quick switching between defined sizes.
But on the other hand I also would like to get the toolbar less crowded.

How do you use it?

robertwidfen:
First of all, thank you for taking the time to work on this and submit a PR. Please don't take this the wrong way, but I have decided to move forward with #549 for this specific issue and its discussion. I will close this PR, but thank you again for your valuable idea! 🤗

PR #549, opened 2026-06-19
refactor: annotation size factor editing into toolbar

robertwidfen:
This was inspired by PR #548 by FedotCompot.

The dialog is now gone and editing is directly possible in the
toolbar. Default step sizes are adjusted to 0.1 and 1.0 and
minimum to 0.01 and maximum to 99.99.

The GTK default bindings are:
 - left mouse button, mouse wheel and up/down step size is 0.1
 - middle mouse button and page up/down step size is 1.0
 - right mouse button jumps to minimum/maximum
 - holding Shift will switch to 0.01 step size

Leave the input with Escape, Return or by clicking somewhere in the
screenshot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,17 @@ Touchpad: <sup>NEXTRELEASE</sup>
 
 If out of range select custom color <sup>0.21.0</sup>
 
+#### Size Selection <sup>NEXTRELEASE</sup>
+
+In the bottom toolbar the size can be edited directly or by + and - buttons.
+Minimum is 0.1 and maximum to 99.99.
+
+The bindings are:
+- Mouse <kbd>left-button</kbd>, <kbd>wheel</kbd> and key <kbd>up</kbd>/<kbd>down</kbd> step size is 0.1
+- Holding <kbd>Shift</kbd> will switch to 0.01 step size
+- Mouse <kbd>middle-button</kbd> and <kbd>page up</kbd>/<kbd>page down</kbd> step size is 1.0
+- Mouse <kbd>right-button</kbd> jumps to minimum/maximum
+
 #### Tool Selection Shortcuts (configurable) <sup>0.20.0</sup>
 Default single-key shortcuts:
 - <kbd>p</kbd>: Pointer tool
--- a/src/sketch_board.rs
+++ b/src/sketch_board.rs
@@ -795,6 +795,10 @@ impl SketchBoard {
         sender: ComponentSender<Self>,
     ) -> ToolUpdateResult {
         match toolbar_event {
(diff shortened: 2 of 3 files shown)


robertwidfen:
Direct comparison to PR#548

I also wonder if we really need the reset.

image

robertwidfen:
In PR #548 [FedotCompot](https://github.com/FedotCompot) asked: At this point remove the S M L buttons and just use the number as a scale factor directly, no?

I am kind of uncertain - I like the S-M-L for quick switching between defined sizes.
But on the other hand I also like to get the toolbar less crowded.

For the moment I would keep S-M-L (`size`) as it is wired into Satty.  All the tools use it and blur-tool uses the `size` but not the `annotation-size-factor`.  So removing `size` would touch many files ...

```rust
pub struct Style {
    pub color: Color,
    pub size: Size,
    pub fill: bool,
    pub annotation_size_factor: f32,
}

pub enum Size {
    Small = 0,
    #[default]
    Medium = 1,
    Large = 2,
}
```

What do others think?

RobertMueller2:
Regarding reset: I don't think the reset button is needed. When I implemented the dialog, the setting was available as config parm and command line parm only, and it was the first time we exposed it via GUI. It felt right at the time, because it's a technical value and I wasn't sure if it was confusing, so I wanted to provide a way to go back to the value that was configured. It has been around for a bit more than a  year and there haven't been many complaints about it. I think it's safe to say the reset button was overcautious.

Regarding S/M/L: I think ONE setting for size would be good. But personally I'm not yet convinced the spin button on its own is intuitive enough. Let's get this merged for now, it's a big improvement as it is.

In the long run, I'd like to pick up discussion again what might be an intuitive control, as has been discussed in #178. Maybe we can take that as a basis too, although it would need some heavy rebasing now, though.,

robertwidfen:
It is not ready to merge, there is still a potential racing condition with updates.

robertwidfen:
IMHO it is now ready for another review and test by others.

To fix the racing condition I removed the 
```
#[watch]
                set_label: &model.annotation_size_formatted,
```
and `annotation_size*` fields - they were actually not necessary anymore.

The start value is set from config at creation.

RobertMueller2:
> IMHO it is now ready for another review and test by others.
> 
> To fix the racing condition I removed the
> 
> ```
> #[watch]
>                 set_label: &model.annotation_size_formatted,
> ```
> 
> and `annotation_size*` fields - they were actually not necessary anymore.
> 
> The start value is set from config at creation.

good catch, I guess that was needed for the interaction of the value with both the label and the dialog.

RobertMueller2 (approved):
lgtm

PR #554, opened 2026-06-23
feat: annotation-size-factor with Ctrl mod and climb rate

robertwidfen:
- Holding <kbd>Ctrl</kbd> now changes the step size to 1.0.
- Restore set_climb_rate to 0.1 (previously lost in PR #549).

Thanks to PR #284 for the inspiration, but a climb rate of 1.0 is IMHO a bit too much.


RobertMueller2 (approved):
lgtm

PR #555, opened 2026-06-24
feat!: generic key bindings

robertwidfen:
See issue #450.

Changes syntax for [keybinds] section in config file to

    "key" = "tool-or-command"

Also adds:
- bindings for all toolbar buttons
- commands for size selection
- automatic hint generation for all buttons

Removes:
- layout independent key bindings - instead bind what you want
- single letter binding restriction

Old style is still supported but will generate deprecation warnings.

The old code always sent key events into the IME handling before they were processed as key event.  Thus there were two places for bindings, bindings with a modifier and single letter bindings. But IME is for text input, thus it is now only activated for the text tool.

### DIFF
--- a/src/keybindings.rs
+++ b/src/keybindings.rs
@@ -45,6 +45,7 @@ pub enum ShortcutCommand {
     SelectSize(Size),
     FocusAnnotationSizeFactor,
     ToggleFill,
+    ToggleRoundCaps,
 }
 
 impl fmt::Display for ShortcutCommand {
@@ -96,6 +97,7 @@ impl fmt::Display for ShortcutCommand {
             },
             ShortcutCommand::FocusAnnotationSizeFactor => "focus-annotation-size-factor",
             ShortcutCommand::ToggleFill => "toggle-fill",
+            ShortcutCommand::ToggleRoundCaps => "toggle-round-caps",
         };
         write!(f, "{}", name)
     }
@@ -161,7 +163,7 @@ impl FromStr for ShortcutCommand {
             "select-size:large" => Ok(ShortcutCommand::SelectSize(Size::Large)),
             "focus-annotation-size-factor" => Ok(ShortcutCommand::FocusAnnotationSizeFactor),
             "toggle-fill" => Ok(ShortcutCommand::ToggleFill),
-
+            "toggle-round-caps" => Ok(ShortcutCommand::ToggleRoundCaps),
             _ => Err(ParseCommandError),
         }
     }
--- a/src/keybindings.rs
+++ b/src/keybindings.rs
@@ -1,0 +1,355 @@
+use std::collections::HashMap;
+use std::fmt;
+use std::str::FromStr;
+use std::sync::OnceLock;
+
+use relm4::gtk;
+use relm4::gtk::gdk::{Key, ModifierType};
+
+use crate::configuration::{APP_CONFIG, Action};
+use crate::sketch_board::KeyEventMsg;
+use crate::style::Size;
+use crate::tools::Tools;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ActionTrigger {
+    Escape,
+    Enter,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ShortcutCommand {
+    // generic
+    ToggleToolbars,
+    OpenGtkInspector,
+    PanLeft,
+    PanRight,
+    PanUp,
+    PanDown,
+    Zoom(i16),
+    DeleteSelection,
+    RunConfiguredActions(ActionTrigger),
+
+    // top toolbar
+    OriginalScale,
+    FitToWindow,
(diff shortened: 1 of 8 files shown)


src/keybindings.rs
```diff
@@ -0,0 +1,364 @@
...
+            "toggle-fill" => Ok(ShortcutCommand::ToggleFill),
+
+            _ => Err(ParseCommandError),
+        }
+    }
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct ShortcutRegistry {
+    key_bindings: HashMap<String, ShortcutCommand>,
```
RobertMueller2: My gut feeling here is, we shouldn't store them as String, but rather a key/modifier pair using the GTK types.

src/keybindings.rs
```diff
@@ -0,0 +1,364 @@
...
+            None
+        } else {
+            None
+        }
+    }
+
+    pub fn get_binding_for_command(&self, command: ShortcutCommand) -> Option<String> {
+        self.key_bindings.iter().find_map(|(binding, cmd)| {
+            if *cmd == command {
+                Some(Self::format_binding_for_hint(binding))
```
RobertMueller2: I haven't tried it, but there's https://gtk-rs.org/gtk3-rs/stable/0.14/docs/gtk/functions/fn.accelerator_get_label.html , perhaps that could be useful instead of this function.

src/ui/toolbars.rs
```diff
@@ -342,28 +335,32 @@ impl SimpleComponent for ToolsToolbar {
...
+            );
+        }
 
-            let tooltip = if let Some(key) = tool_to_key_map.get(tool) {
-                &format!("{} ({})", display_name, key.to_uppercase())
-            } else {
-                display_name
-            };
-            button.set_tooltip_text(Some(tooltip));
+        #[rustfmt::skip]
```
RobertMueller2: I'm not a huge fan of skipping anything, but I can definitely understand why you did it. Perhaps we just need a better rustfmt config at some point in the future, e.g. with longer lines.

src/keybindings.rs
```diff
@@ -0,0 +1,364 @@
...
+            ShortcutCommand::RunConfiguredActions(ActionTrigger::Enter),
+        );
+        registry.add_key_binding(
+            "KP_Enter",
+            ShortcutCommand::RunConfiguredActions(ActionTrigger::Enter),
+        );
+
+        // top toolbar
+        registry.add_key_binding("<Control>1", ShortcutCommand::OriginalScale);
+        registry.add_key_binding("<Control>2", ShortcutCommand::FitToWindow);
```
RobertMueller2: In #338 I had the idea to add a few more shortcuts for specific zoom levels, e.g.
ctrl+2 - 200%, ctrl+3 - 300%, ctrl+4 - 400%, ctrl+5 - 500%
ctrl+shift+2 - 50%, ctrl+shift+3 - 25% ....

This can be done later, but I'd prefer not to block it. Of course it's debatable if a configurable binding can block anything, but I'd least I'd like that one to be considered experimental.

Or use an alternative, e.g. ctrl+shift+1, or ctrl+0 (although the latter is really far away).

RobertMueller2:
Good work, code looks good to me with just a few tiny inline remarks. Not that many for such a big change. What I tested worked. Difficult to try out everything :smile:

IME still works as far as I can tell. But I'm not the greatest expert for it. I'm not sure if I should say "it still works", perhaps somebody else sees this and can confirm this.

src/keybindings.rs
robertwidfen: I also considered <kbd>Ctrl+0</kbd> but most of the time I have my left hand on keyboard and the right on the mouse and thus it is too hard to reach. 😉

src/keybindings.rs
robertwidfen: Indeed much nicer, thanks. Just 5 lines of code.

src/ui/toolbars.rs
robertwidfen: Me neither, but it looked so ugly - it is no the line length - 372 is longer.

Image

src/keybindings.rs
robertwidfen: Should not be complicated, they are parsed anyway during validation, but I am not sure what value it would bring beside a minor speedup during lookup. If GTK changes the format it would affect may lines either way. During testing and debugging it was nice to have a readable string.

robertwidfen: It is now with the types and some other changes.

src/ui/toolbars.rs
robertwidfen: I "aliased" by `type SC = ShortcutCommand;` and it looks mostly nice.

src/keybindings.rs
RobertMueller2: I'll add a new comment for this, probably should be in config.toml anyway.

RobertMueller2:
lgtm :confetti_ball:

PR #556, opened 2026-06-25
ci: add aarch64 (arm64) Linux release build

MAY4VFX:
## What

Adds an **aarch64 (arm64) Linux** target to the `release` workflow, so each release also ships `satty-aarch64-unknown-linux-gnu.tar.gz` alongside the existing x86_64 tarball.

## Why

There are currently no prebuilt aarch64 binaries. Users on arm64 Linux (Asahi/Apple Silicon Linux, Raspberry Pi, Pinebook Pro, arm64 servers/VMs, etc.) have to build from source, which is slow on these devices.

## How

The existing build runs inside `ghcr.io/gtk-rs/gtk4-rs/gtk4:latest`, which is **amd64-only**, so it can't simply be reused on an arm64 runner. Instead the aarch64 leg:

- runs natively on GitHub's free **`ubuntu-24.04-arm`** runner (no cross-compilation),
- runs **without a container** (the container is now selected per-matrix-entry via `container: ${{ matrix.container }}`; empty string = run on host),
- installs GTK4/libadwaita via `apt` (`libgtk-4-dev libadwaita-1-dev`) instead of the Fedora container's `yum`.

The x86_64 leg is unchanged. Diff is +16/-5.

## Tested

Built green on a fork — both legs and the flatpak job succeed, and the aarch64 tarball is produced and attached to the release:

https://github.com/MAY4VFX/Satty/releases/tag/v0.99.0

The resulting binary runs correctly on a Pinebook Pro (RK3399, ARMv8.0) under Debian 13 / niri.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,18 +16,30 @@ jobs:
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
             archive: tar.gz
+            container: ghcr.io/gtk-rs/gtk4-rs/gtk4:latest
+          - os: ubuntu-24.04-arm
+            target: aarch64-unknown-linux-gnu
+            archive: tar.gz
+            container: ""
 
     runs-on: ${{ matrix.os }}
-    container:
-      image: ghcr.io/gtk-rs/gtk4-rs/gtk4:latest
+    # x86_64 builds in the gtk4-rs Fedora container (yum); aarch64 runs natively
+    # on GitHub's free arm64 runner with GTK installed via apt (empty container).
+    container: ${{ matrix.container }}
 
     permissions:
       contents: write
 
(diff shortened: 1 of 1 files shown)


robertwidfen:
LGTM - Claude did a good job 🤣

But I am not into this workflow stuff - so another 👁️ by @RobertMueller2 would be good.

RobertMueller2:
yeah, I'll have a look soonish. Should even be able to try the generated package on a chromebook ;)

EDIT: which has a Bookworm Debian, which is too old (libc6 version). Need to figure out a different way to test the package. Or upgrade the Debian version, but not today.

RobertMueller2:
lgtm, thanks!

PR #558, opened 2026-06-29
feat: make clear-all undoable as an atomic history action

### LINKED ISSUE #524: feature: Make reset undoable as a single history entry
### Description

Currently, reset removes all current drawables by moving them into the redo stack.

This makes the resulting history behavior unintuitive from a user perspective:

  ```text
  draw A
  draw B
  reset
  Ctrl+Z -> nothing is restored
  Ctrl+Y -> restores one drawable
  Ctrl+Y -> restores another drawable
```

In visual editors and other applications with undo/redo history, I would expect reset/clear-all to behave as a single undoable action:

```
  draw A
  draw B
  reset
  Ctrl+Z -> restores A and B
  Ctrl+Y -> applies reset again
```

One possible implementation would be to represent reset explicitly in the history, for example:

```rust
  enum HistoryEntry {
      Drawable(Box<dyn Drawable>),
      Reset(Vec<Box<dyn Drawable>>),
  }
```

Then reset could be stored as an atomic history entry rather than being modeled as many individual undo operations pushed into `redo_stack`.

This would make `Ctrl+Z` after reset match the visible last user action more directly, while still allowing `Ctrl+Y` to re-apply the reset if needed.

I’d like to discuss whether this behavior matches the intended UX for Satty before opening an implementation PR.

sonnymilton:
## Summary

  Closes #524 

  This changes the current reset/clear-all behavior so it is represented as an atomic history entry instead of moving all drawables directly into the redo stack.

  After this change:

  ```text
  draw A
  draw B
  clear all
  Ctrl+Z -> restores A and B
  Ctrl+Y -> clears all again
```

  The toolbar action is also renamed from Reset to Clear all to better match the user-facing behavior and the distinction discussed in the issue.

  ## Implementation

  - Adds `HistoryEntry::ClearAll(Vec<Box<dyn Drawable>>)` for atomic clear-all history.
  - Keeps visible drawables separate from undo/redo history.
  - Updates undo/redo handling so clear-all can be undone and redone as one action.
  - Renames the renderer action from `reset()` to `clear_all().`
  - Renames the toolbar event and tooltip from `Reset` to `Clear all`.

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -50,7 +50,8 @@ pub struct FemtoVgAreaMut {
     scale_factor: f32,
     offset: Vec2D,
     drawables: Vec<Box<dyn Drawable>>,
-    redo_stack: Vec<Box<dyn Drawable>>,
+    undo_stack: Vec<HistoryEntry>,
+    redo_stack: Vec<HistoryEntry>,
     zoom_scale: f32,
     last_scale: f32,
     pointer_offset: Vec2D,
@@ -58,6 +59,11 @@ pub struct FemtoVgAreaMut {
     drag_offset: Vec2D,
     is_drag: bool,
     is_reset: bool,
+}
+
+enum HistoryEntry {
+    Drawable(Box<dyn Drawable>),
+    ClearAll(Vec<Box<dyn Drawable>>),
 }
 
 #[glib::object_subclass]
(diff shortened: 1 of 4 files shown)


RobertMueller2:
My initial thought was, we'd keep the existing behaviour as well. But I suppose that has always been a workaround, so it's probably better that we don't. Looks good at first glance, need to test it a bit more.

RobertMueller2:
LGTM, thanks for the PR.

Issue #463, opened 2026-03-18
Followup: GUI revamp


RobertMueller2:
#49 started as a specific bug report, but ended up having a number of suggestions for a GUI revamp. Now that #49 is closed, use this issue as a followup for any GUI revamp related discussions.

PR #561, opened 2026-06-30
feat: tool dismiss/commit buttons in style toolbar

### LINKED ISSUE #472: Leftover comment after PR #357 review
This commented section should be removed, I forgot that before merging.

https://github.com/Satty-org/Satty/blob/590253c8bba01f5c559808dd432cbb80bc4bd7c8/src/sketch_board.rs#L760

RobertMueller2:
Adds tool dismiss/commit buttons. Introduces handle_dismissed method for tools.

The buttons only really work for crop and text right now, but should be useful once we get editable components.

Change text so Esc used handle_dismissed (README says so anyway), enter is already available for commit.

https://github.com/user-attachments/assets/88035651-4847-45e4-9a86-1445a65bc7e0

because it annoyed me while editing:
Closes: #472

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -74,6 +74,8 @@ fn main() -> Result<(), io::Error> {
             "page-fit-regular",
             "resize-large-regular",
             "arrow-counterclockwise-regular",
+            "dismiss-regular",
+            "checkmark-regular",
         ],
     );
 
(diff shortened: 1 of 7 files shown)


robertwidfen:
LGTM.

My only concern is that Satty gets wider with each new toolbar button.

For a moment I was thinking a click on the active tool could do it - but it is not clear if that should do a commit or dismiss.  The buttons on the other hand are clear.

RobertMueller2:
I entirely agree with you regarding the concern, I don't think I have short term solution for it :) We do have #463 to discuss this, though.

Issue #470, opened 2026-03-28
Right click to finish crop editing also triggers action


RobertMueller2:
### Version

a14c1d16e32641d688de4203f78b481a4b7a5044

### Description

Commit a14c1d16e32641d688de4203f78b481a4b7a5044 introduced right click to clear crop (while editing crop).

If there's also actions-on-right-click, both are executed, this shouldn't happen.

Alternatively, Right-Click was a poor choice for this.

duguyipiao:
Is it possible to make Right-Click to clear crop optional?

My workflow was running satty for the full screen and then cropping the selected region (like flameshot). I use `actions-on-right-click = ["save-to-clipboard", "save-to-file", "exit"]` to save the selected region. The latest version completely break the workflow.

I second that Right-Click was a poor choice for clearing crop.

RobertMueller2:
> Is it possible to make Right-Click to clear crop optional?

I don't think we'll add an option just for that. I'm sorry, but I don't think there's any short term relief for your issue.

But in the long run, after #450 is implemented, I think that tool specific and editing specific (e.g. commit and discard) command mappings should become editable also.

RobertMueller2:
FYI, we merged #562 which maps the crop cancellation to ctrl+right click. So your issue should be resolved with the next release.

PR #562, opened 2026-06-30
fix!: unclash right mouse button in crop tool

### LINKED ISSUE #560: --actions-on-right-click save-to-clipboard no longer works in crop mode
### Version

0.21.1.r15.g770a7e3-1

### Description

I use the following command to run satty:

```fish
grim \
    -c \
    -o (swaymsg -t get_outputs | jq -r '.[] | select(.focused) | .name') \
    -t png \
    - | 
satty \
	--filename - \
	--fullscreen \
	--early-exit all \
	--initial-tool crop \
	--disable-notifications \
	--actions-on-right-click save-to-clipboard \
	--save-after-copy \
	--actions-on-escape exit \
        --no-window-decoration \
        --copy-command wl-copy \
        --corner-roundness 0 \
        --focus-toggles-toolbars \
	--output-filename ~/Pictures/satty/(date +"%Y-%m-%d_%H-%M-%S").png
```

After some update, satty stopped working like it used to. Previously, right clicking inside the crop area would save the image to my clipboard and to a file (cropped if I click inside the crop area, full screen if I click outside of it). Now, it seems to only reset my crop, no matter where i right click.

I have to manually click the copy button on the toolbar instead of right clicking for it to copy the image to the clipboard, or manually select the mouse tool and then it works as it previously did.

I have experimented with different options, but there doesn't seem to be a difference.
```
--actions-on-right-click save-to-clipboard,save-to-file,exit
```
```
--actions-on-right-click save-to-clipboard \
--actions-on-right-click save-to-file \
--actions-on-right-click exit \
```

`satty --help` has this:
```
--save-after-copy
          After copying the screenshot, save it to a file as well Preferably use the `action_on_copy` option instead
```
However, I can't figure out what `action_on_copy` is. `--action-on-copy` doesn't exist or is not documented properly.

RobertMueller2:
Closes: #560
Addresses complaint in #470

When using right mouse button for actions, in crop it may be counter-intuitive that right click resets crop.

For now, add Ctrl modifier to the right click. This is experimental because in the future tool related bindings/shortcut may become configurable.

Also add Ctrl+Left click to finish crop editing.

PR builds on #561, because that one offers a mouse-only way to apply and reset crop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Default single-key shortcuts:
 ### Tool Modifiers and Keys
 
 Crop:
-- Press <kbd>Esc</kbd> or right mouse button while editing to reset crop altogether <sup>0.21.0</sup>.
+- Press <kbd>Esc</kbd> or <kbd>Ctrl</kbd>+right mouse<sup>NEXTRELEASE</sup> <sup>experimental</sup> button while editing to reset crop altogether <sup>0.21.0</sup>.
 - Press <kbd>Enter</kbd> while editing to finish editing crop and keep the crop area active <sup>0.21.0</sup>.
 - Left click crop area when tool is active but not editing to resume editing<sup>0.21.0</sup>.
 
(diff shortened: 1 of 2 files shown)


robertwidfen:
LGTM

I like if binding do not "overlap" and might trigger another action by accident.

PR #565, opened 2026-07-02
chore: update rust-overlay flake input to fix the dev shell build

### LINKED ISSUE #564: [Bug]: nix dev shell can no longer build the project (pinned rust-overlay provides rustc 1.87, dependencies require 1.92)
### Version

commit 770a7e35f1d78bf0dcb926f64303068ca21f43c8

### Description

The `flake.lock` pins `rust-overlay` at a revision from 2025-05-25 which provides rustc 1.87. Meanwhile the dependencies in `Cargo.lock` declare higher MSRVs (relm4 0.10.1, relm4-icons-build 0.11 and gvdb 0.10 require 1.92; femtovg 0.25, image 0.25.10 and libloading 0.9 require 1.88), so building inside the dev shell fails out of the box:

```
$ nix develop --command cargo build
error: rustc 1.87.0 is not supported by the following packages:
  femtovg@0.25.0 requires rustc 1.88
  gvdb@0.10.0 requires rustc 1.92
  relm4@0.10.1 requires rustc 1.92
  ...
```

This only affects the flake dev shell (CI and distro packages build with their own toolchains), but the dev shell is the documented entry point for contributors on NixOS (see also #262), and right now it cannot build the project at all.

The fix is a one-liner: `nix flake update rust-overlay`. I have a branch ready and can open a PR.

For reference, PR #563 was built with `nix develop --override-input rust-overlay github:oxalica/rust-overlay` as a workaround.

rokokol:
Closes #564 

`nix flake update rust-overlay` — the pinned revision provides rustc 1.87 while the dependencies in `Cargo.lock` require up to rustc 1.92, so `nix develop --command cargo build` fails out of the box (see the issue for the full error).

With this update the dev shell provides rustc 1.96.1 and `cargo build`, `cargo clippy` and `cargo test` all work again. `nixpkgs` is left untouched to keep the change minimal.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -43,11 +43,11 @@
         "nixpkgs": "nixpkgs_2"
       },
       "locked": {
-        "lastModified": 1748140821,
-        "narHash": "sha256-GZcjWLQtDifSYMd1ueLDmuVTcQQdD5mONIBTqABooOk=",
+        "lastModified": 1782962170,
+        "narHash": "sha256-HsaK9yDDypMB2TPxAYz0mPKtmYhgNwFQa72b3Ko6kRE=",
         "owner": "oxalica",
         "repo": "rust-overlay",
-        "rev": "476b2ba7dc99ddbf70b1f45357dbbdbdbdfb4422",
+        "rev": "e3fa5cf86b93914b8f312b2a1ca14fbb139c655c",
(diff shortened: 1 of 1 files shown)


RobertMueller2:
LGTM, thanks :)

PR #571, opened 2026-07-03
chore(deps): update fontconfig to 0.11 (incompatible)

RobertMueller2:
`Fontconfig.find()` now is a Result rather than an Option, requiring a minor code change.


robertwidfen:
LGTM

PR #574, opened 2026-07-04
feat: toggle text outline in inverted color with Alt

### LINKED ISSUE #573: Text tool: option to draw text with a contrasting outline
### Description

Problem

When placing text annotations over a screenshot, the text color can blend into the background (e.g. dark text over a dark area, or a bright color over a busy image), making it hard to read. Today the only workaround is to manually pick a color that happens to contrast with whatever is underneath the text

rokokol:
Closes #573  

# What

Adds an optional outline to the text tool. While editing text, tapping <kbd>Alt</kbd> toggles an outline drawn in the inverted text color (255 - channel per RGB, alpha preserved). The outline is preserved when the text is committed.

Adds an optional outline to the text tool. While editing text, tapping <kbd>Alt</kbd> toggles an outline drawn in the inverted text color (255 - channel per RGB, alpha preserved). The outline is preserved when the text is committed.

https://github.com/user-attachments/assets/a43b4b83-dc34-4fe5-bddd-1351b5b005b8

## Why

Text can blend into the background and become unreadable. An inverted-color outline always contrasts with the fill, keeping annotations legible over any background, with no new color setting to configure.

## How

- Color::inverted() in style.rs — per-channel RGB inversion, alpha kept.
- In Text::draw, when the outline is enabled each line is stroked with the inverted color (line width ×2) underneath the fill, so the visible border wraps the glyphs. The outline flag lives on Text, so it is carried into the committed drawable via clone_box.
- Toggling uses a genuine Alt tap rather than the raw keydown: handle_key_event marks a tap candidate on Alt_L/Alt_R and clears it on any other key; handle_key_release_event flips outline only if the candidate survived. This prevents Alt-based combos (Ctrl+Alt+Arrow to move text) and key auto-repeat from flipping the state.

## Docs

- README.md: added the shortcut under the Text tool section with a NEXTRELEASE marker.

## Testing

- make fix (cargo fmt + clippy -D warnings) — clean.
- cargo build — clean.
- tested on NixOS with Hyprland.

## Notes

No new dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ Text:
 - <kbd>Ctrl+X</kbd> to cut selected text to clipboard. <sup>0.20.1</sup>
 - <kbd>Ctrl+V</kbd> to paste text from clipboard. <sup>0.20.1</sup>
 - <kbd>Alt+Ctrl</kbd> with <kbd>Left</kbd> or <kbd>Right</kbd> or <kbd>Up</kbd> or <kbd>Down</kbd> to move the text. Use <kbd>Alt+Ctrl+Shift</kbd> with arrow keys to nudge the text. <sup>0.20.1</sup>
-- Press <kbd>Alt</kbd> to toggle outlining the text with the inverted text color. <sup>NEXTRELEASE</sup>
+- Press <kbd>Alt</kbd> to toggle outlining the text with the inverted text color. <sup>experimental</sup> <sup>NEXTRELEASE</sup>
 
 Marker:
 - Hold <kbd>Alt</kbd> to get extra ring. <sup>NEXTRELEASE</sup>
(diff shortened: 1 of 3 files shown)


RobertMueller2:
Looks good, nice addition.

We may revamp the way tool styles, secondary modes, tool specific actions etc are accessed in the future, ideally even allow configuring keybindings for this. Will definitely be after #555. In order to allow that, it'd be nice if you could flag the Alt toggle in the README as Experimental. Thanks :)

I also have a bonus idea, feel free to reject it, though, we can always do that later. :wink: How about there are different outline modes:

```
enum OutlineMode {
  #[default]
  None,
  Inverted,
  Contrast
}
```

and hitting alt cycles through them?

`Inverted` is what you just implemented, Contrast picks black or white based on Luminance?

rokokol:
Done! My goal was to keep everything as simple and concise as possible, without any extra interface buttons or unnecessary settings, so that the app would remain minimalistic and straightforward :Ъ

RobertMueller2:
lgtm, thanks for the PR :)

PR #576, opened 2026-07-04
 feat: drop-shadow text effect + gamma-corrected contrast

rokokol:
Follow-up to #574 (text outline). Two small, independent improvements — one commit each:

1. Drop-shadow effect <sup>3794dc2</sup>
  Generalizes the Alt-cycled outline into a small TextEffect set. Each variant yields a Decoration describing what's drawn behind the
fill (an outline stroke, or an offset shadow duplicate), so new effects are one enum variant + one match arm. New cycle: none → inverted outline → contrast outline → drop shadow.

2. Gamma-corrected contrast <sup>e8f623a</sup>
  Color::contrast() now uses WCAG relative luminance (sRGB channels gamma-decoded to linear light, then perceptually weighted), with the L = 0.179 black/white crossover, instead of a raw YIQ sum. More legible choice on mid-tones. Also adds a small Color::with_alpha helper.

README.md updated (effect list, still flagged experimental). make fix clean; both commits build standalone.


RobertMueller2:
seems like a logical addition now, thanks for the PR :)

Issue #309, opened 2025-10-16
Feature idea: add fallback fonts automatically


RobertMueller2:
Followup for #308 and #298 

The idea is to explore using an advanced text engine like Pango or Cosmic Text to determine fallback fonts for the entered text, these could be used for femtovg rendering.

Femtovg itself considers fallback fonts, but it depends on which fonts have been added.
https://github.com/femtovg/femtovg/blob/master/examples/text.rs

While #298 provides a configuration option for this, the convenience isn't great.

https://docs.gtk.org/Pango/method.Fontset.get_font.html
https://github.com/femtovg/femtovg/blob/master/examples/external_text.rs

An important point, next to convenience, should be performance. The font loading mechanism shouldn't lead to additional sluggishness.

gabm:
Just a thought: we could also consider contributing to femtovg, if thats a better place for these things... afterall we could give something back!

RobertMueller2:
Sure, when we get to that point, we can ask. But since the external text example does make use of cosmic_text, I'm not sure if the project will want to have any more sophisticated fallback font loading. I'm personally not sure if I'll ever have the knowledge of making this super useful, then again, it can probably be researched in both the above.

Also, different thought, we're already using fontconfig, I think we may even be able to get fallback fonts that way.

PR #577, opened 2026-07-04
feat: render emoji in the text tool

### LINKED ISSUE #536: Support for emojis in text tool
### Description

Would be nice.

IMHO this is a limitation of femtovg - I could not get it working - played with fallback fonts and also monochrome Noto Emoji.

Other tools like flameshot and swappy work out of the box.

Was anybody able to get it working?

rokokol:
Fixes #536
## Summary

Enable color emoji rendering in the text tool. An installed color emoji font (e.g. Noto Color Emoji) is automatically picked up as the last fallback, so emoji work out of the box without any user configuration.

Text outlines (Alt toggle) are not applied to emoji — they are rendered as-is using their native bitmap colors, while regular text glyphs continue to receive the outline as before.

image

## Problem

Emoji typed into the text tool were silently not rendered — only regular glyphs appeared.

Two things were missing:

1. **No emoji font in the fallback chain.** The font loading code only loaded the user-configured primary font and explicit fallbacks. Without a color emoji font in the stack, femtovg had no glyphs to fall back on for emoji codepoints.

2. **PNG decoder not compiled in.** femtovg supports color emoji via its `image-loading` feature (enabled by default), which decodes PNG glyph bitmaps embedded in color fonts (CBDT/sbix tables). However, femtovg declares `image = { default-features = false }` — meaning the `png` decoder is not included. The call to `image::load_from_memory_with_format(..., Png)` silently failed, and emoji glyphs were skipped entirely.

## Changes

- **`src/femtovg_area/imp.rs`**: Append the fontconfig generic `"emoji"` family as the last fallback font. fontconfig resolves this to the installed color emoji font (e.g. Noto Color Emoji). Skipped gracefully if already loaded or unavailable.
- **`Cargo.toml`**: Add `image` with `features = ["png"]` as a direct dependency. This is not a new dependency — `image` is already pulled in transitively through femtovg. The explicit entry activates the `png` feature via Cargo's feature unification, enabling the PNG decoder that femtovg needs for color glyph bitmaps.
- **`README.md`**, **`config.toml`**: Document that emoji fonts are picked up automatically.

## Outline behavior

femtovg internally separates color glyphs (bitmap emoji) from path-based glyphs during text rendering. When `stroke_text` is called, color glyphs are rendered with `line_width = 0` — the stroke outline is never applied to them. They appear using their native bitmap colors regardless of the outline paint. This is handled entirely within femtovg and requires no special-casing in Satty.

## Testing

- Typed emoji (e.g. 😀🎉🔥) in the text tool → rendered correctly as color bitmaps
- Mixed text + emoji on the same line → both render correctly
- Outline mode (Alt toggle) → outline applied to text glyphs only, emoji remain unaffected

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -262,6 +262,19 @@ impl FemtoVGArea {
             }
         }
 
+        // Append a color emoji font as the last fallback so emoji render out of the box
+        // (femtovg's default image-loading feature draws their bitmap/COLR glyphs).
+        // fontconfig resolves the generic "emoji" family to the installed color emoji
+        // font, e.g. Noto Color Emoji. It is skipped if already loaded or unavailable.
+        match load_font("emoji", None) {
+            Ok(id) => {
+                loaded_fonts.push(id);
+            }
+            Err(e) => {
+                eprintln!("Emoji font: {}", e);
+            }
+        }
+
         Ok((text_context, loaded_fonts))
     }
 
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -266,6 +266,7 @@ impl FemtoVGArea {
         // (femtovg's default image-loading feature draws their bitmap/COLR glyphs).
         // fontconfig resolves the generic "emoji" family to the installed color emoji
         // font, e.g. Noto Color Emoji. It is skipped if already loaded or unavailable.
+        // reconsider with #309
         match load_font("emoji", None) {
             Ok(id) => {
                 loaded_fonts.push(id);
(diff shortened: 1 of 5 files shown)


src/femtovg_area/imp.rs
```diff
@@ -262,6 +262,19 @@ impl FemtoVGArea {
             }
         }
 
+        // Append a color emoji font as the last fallback so emoji render out of the box
```
RobertMueller2: Please let's add a reference here to #309, if we ever implement that, this may have to be changed or removed.

E.g. "reconsider with #309" or so.

RobertMueller2:
Good find and thanks for investigating. I only have that inline remark. :)

src/femtovg_area/imp.rs
rokokol: Done. I was thinking about that and forgot to mention it 😅

PR #583, opened 2026-07-12
feat: add notification-thumbnail config/cmdline option to allow sending notification with image thumbnail rather than app icon

### LINKED ISSUE #426: [Feature Request] Include the screenshot image saved as icon or miniature in notifications
### Description

Hi. First of all, thank you very much for all the hard work to develop and maintain this amazing tool.

I'm wandering if it would be possible to use the screenshot finally saved within the notification instead of the Satty icon just as hyprshot does for example. I think it would be a very cool and nice touch of "elegance".

Thanks and keep up the good work!

RobertMueller2:
Closes: #426 

Also gets rid of redundant `!APP_CONFIG.read().disable_notifications()` by placing that one centrally in `log_result`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ default-hide-toolbars = false
 focus-toggles-toolbars = false
 # Fill shapes by default (since 0.20.0)
 default-fill-shapes = false
-# Round caps for `arrow` and `line` tools (always true for `brush` tool)
+# Round caps for arrow and line tools
 default-round-caps = true
 # The primary highlighter to use, the other is accessible by holding CTRL at the start of a highlight [possible values: block, freehand]
 primary-highlighter = "block"
@@ -219,8 +219,8 @@ title = "Satty"
 # experimental feature (0.21.0): set app_id, note this has to match D-Bus well-known name format, otherwise GTK does not accept it.
 app-id = "org.satty.satty"
 # experimental feature (NEXTRELEASE): show thumbnail as notifcation icon
-# notification-thumbnail = "thumbnail-file--icon"
-notification-thumbnail = "app-icon"
+# notification-thumbnail = "app-icon"
+notification-thumbnail = "screenshot"
 
 # Tool selection keyboard shortcuts (since 0.20.0)
 [keybinds]
@@ -360,7 +360,7 @@ Options:
       --app-id <APP_ID>
           Experimental feature (0.21.0): Set toplevel app_id. Note that this has to match D-Bus well known name format, otherwise GTK does not accept it
       --notification-thumbnail <NOTIFICATION_THUMBNAIL>
-          Experimental feature (NEXTRELEASE): use preview thumbnail in notifications where available [possible values: app-icon, thumbnail-file-icon]
+          Experimental feature (NEXTRELEASE): use preview thumbnail in notifications where available [possible values: screenshot, app-icon]
       --right-click-copy
           Right click to copy. Preferably use the `action_on_right_click` option instead
       --action-on-enter <ACTION_ON_ENTER>
--- a/README.md
+++ b/README.md
@@ -218,6 +218,9 @@ input-scale = 1.0
 title = "Satty"
 # experimental feature (0.21.0): set app_id, note this has to match D-Bus well-known name format, otherwise GTK does not accept it.
 app-id = "org.satty.satty"
+# experimental feature (NEXTRELEASE): show thumbnail as notifcation icon
+# notification-thumbnail = "thumbnail-file--icon"
+notification-thumbnail = "app-icon"
 
 # Tool selection keyboard shortcuts (since 0.20.0)
 [keybinds]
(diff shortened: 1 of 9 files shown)


RobertMueller2:
The clippy warning happen in main, too. I'll open a separate PR for that.

src/notification.rs
```diff
@@ -1,35 +1,100 @@
...
+use relm4::gtk::gdk_pixbuf::{InterpType, Pixbuf};
+use relm4::gtk::prelude::Cast;
+use relm4::gtk::{IconLookupFlags, IconTheme, TextDirection, gio};
+use satty_cli::command_line::NotificationThumbnail;
+
+pub static NOTIFICATION_THUMBNAIL_PATH: LazyLock<PathBuf> = LazyLock::new(|| {
+    std::env::temp_dir().join(format!(
+        "satty-notification-thumbnail-{}.png",
+        std::process::id()
+    ))
```
robertwidfen: This is unsafe, other malicious apps might guess and create the same file - on a single user machine likely less critical.

Also using a single filename means it will overwrite this with each notification. It is only living a short time here, so it is no issue IMHO.

The right way is to use the tempfile crate - at least this is what I use in another tool. 😉

src/notification.rs
```diff
@@ -1,35 +1,100 @@
...
+            }
+        }
+    };
+
+    let icon = match pixbuf {
+        // glib doesn't support image-data at this point
+        /*
+        Some(p) if notification_icon_kind == NotificationThumbnail::ThumbnailIcon => {
+            Some(p.upcast::<Icon>())
+        }*/
```
robertwidfen: Maybe delete this?

Too bad that we need the tempfile and cannot use the pixbuf.

robertwidfen (commented):
Works nicely - just two comments below.

Luckily, I've just set up Noctalia, so the thumbnails are showing up nicely.

src/notification.rs
RobertMueller2: > This is unsafe, other malicious apps might guess and create the same file - on a single user machine likely less critical.

Well, what could happen? If we can write the file, we overwrite it. If we can't, we get an error and the notification shows satty's app icon. If somebody manages to sneak in their own icon into the notification, gg to them  :laughing: 

A malformed image could be an issue, I suppose.

> Also using a single filename means it will overwrite this with each notification. It is only living a short time here, so it is no issue IMHO.

A notification could display a "wrong" icon, or if the timing is very bad, perhaps the image corrupts.

It was the easiest solution for cleanup, because you can't delete it right away or the notification can fail due to the file not existing, that is something I actually observed.

> The right way is to use the tempfile crate - at least this is what I use in another tool. 😉

Well, the problem is the cleanup. It's not like I haven't thought about this. I can't delete the file right away, as mentioned before. Both tempfile and NamedTempfile delete the file too early, unless we store all of them until they can be deleted.

With that in mind, having a single file name that can be cleaned up at program exit seemed like it wasn't the worst thing to do here.

src/notification.rs
RobertMueller2: indeed, just needed to make sure there's a reference in the PR ;)

cli/src/command_line.rs
```diff
@@ -215,6 +219,22 @@ pub enum EarlyExitTriggers {
...
+// options are: use different notification library or implement manually. For now, we
+// just offer image-path (which uses a temporary file).
+#[derive(Debug, Clone, Copy, Default, ValueEnum, Deserialize, PartialEq, Eq)]
+#[value(rename_all = "kebab-case")]
+#[serde(rename_all = "kebab-case")]
+pub enum NotificationThumbnail {
+    #[default]
+    AppIcon,
+    ThumbnailFileIcon,
+    //ThumbnailIcon
```
robertwidfen: Also just use "Screenshot" as enum as "ThumbnailIcon" and "ThumbnailFileIcon" are just implementation details which should not go into config/args.

I also vote for "Screenshot" as the default, so users will get this nice feature without config change.

config.toml
```diff
@@ -78,6 +78,9 @@ input-scale = 1.0
 title = "Satty"
 # experimental feature (0.21.0): set app_id, note this has to match D-Bus well-known name format, otherwise GTK does not accept it.
 app-id = "org.satty.satty"
+# experimental feature (NEXTRELEASE): show thumbnail as notifcation icon
+# notification-thumbnail = "thumbnail-file--icon"
+notification-thumbnail = "app-icon"
```
robertwidfen: As above "thumbnail" as default.

README.md
```diff
@@ -218,6 +218,9 @@ input-scale = 1.0
 title = "Satty"
 # experimental feature (0.21.0): set app_id, note this has to match D-Bus well-known name format, otherwise GTK does not accept it.
 app-id = "org.satty.satty"
+# experimental feature (NEXTRELEASE): show thumbnail as notifcation icon
+# notification-thumbnail = "thumbnail-file--icon"
+notification-thumbnail = "app-icon"
```
robertwidfen: As above "thumbnail" as default.

README.md
```diff
@@ -356,6 +359,8 @@ Options:
           Experimental feature (0.21.0): Set window title
       --app-id <APP_ID>
           Experimental feature (0.21.0): Set toplevel app_id. Note that this has to match D-Bus well known name format, otherwise GTK does not accept it
+      --notification-thumbnail <NOTIFICATION_THUMBNAIL>
+          Experimental feature (NEXTRELEASE): use preview thumbnail in notifications where available [possible values: app-icon, thumbnail-file-icon]
```
robertwidfen: As above "thumbnail" as default.

robertwidfen (commented):
LGTM - just change the enum and update config & README and then merge.

robertwidfen:
Rethinking this just one time more an it should be:
```
--notification-thumbnail=screenshot
--notification-thumbnail=app-icon
```

robertwidfen (commented):
--notification-thumbnail=screenshot

PR #592, opened 2026-07-21
feat: default rounded brush lines and round cap toggle button

### LINKED ISSUE #590: fix & enhancement: rounded brush lines
### Description

Hello,
I use satty everyday, that's my main annotation tool and I love it.
I always had a problem with brush style: it seemed square-like and ragged: 

Image

I supposed, that annotation tools users expect, that brush line will be rounded, so I found a place in a style.rs file with `From<Style>` implementation for `Paint` type:

```rs
...
impl From<Style> for Paint {
    fn from(value: Style) -> Self {
        Paint::default()
            .with_anti_alias(true)
            .with_font_size(value.size.to_text_size(value.annotation_size_factor) as f32)
            .with_color(value.color.into())
            .with_line_width(value.size.to_line_width(value.annotation_size_factor))
    }
}
...
```

and added just two lines in my fork for round `Paint` initialization:
```rs
...
impl From<Style> for Paint {
    fn from(value: Style) -> Self {
        Paint::default()
            .with_anti_alias(true)
            .with_font_size(value.size.to_text_size(value.annotation_size_factor) as f32)
            .with_color(value.color.into())
            .with_line_join(LineJoin::Round) // round join
            .with_line_cap(LineCap::Round) // round line cap
            .with_line_width(value.size.to_line_width(value.annotation_size_factor))
    }
}
...
```

So, as a result: 

Image

However, sometimes there are some sharp "pacman"-like edges: 

Image

imizgun:
Closes #590 

This PR adds `.with_line_cap(LineCap::Round)` for `brush` tool as default drawing behavior, adds new button to the toolbar (`Toggle Round Caps`) next to the `fill` function. This button toggles round caps for `arrow` and `line` tools.

https://github.com/user-attachments/assets/36aaf000-e818-4db3-8a28-247545662b27

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ default-hide-toolbars = false
 focus-toggles-toolbars = false
 # Fill shapes by default (since 0.20.0)
 default-fill-shapes = false
+# Round caps for `arrow` and `line` tools (always true for `brush` tool)
+default-round-caps = true
 # The primary highlighter to use, the other is accessible by holding CTRL at the start of a highlight [possible values: block, freehand]
 primary-highlighter = "block"
 # Disable notifications
@@ -307,7 +309,7 @@ Options:
       --corner-roundness <CORNER_ROUNDNESS>
           Draw corners of rectangles round if the value is greater than 0 (Defaults to 12) (0 disables rounded corners)
       --initial-tool <TOOL>
-          Select the tool on startup [aliases: --init-tool] [possible values: pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, highlight, brush]
+          Select the tool on startup [alias: --init-tool] [possible values: pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, highlight, brush]
       --copy-command <COPY_COMMAND>
           Configure the command to be called on copy, for example `wl-copy`
       --annotation-size-factor <ANNOTATION_SIZE_FACTOR>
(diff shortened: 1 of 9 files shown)


RobertMueller2:
Thanks for the PR. Now that we got this, I'm kinda wondering if we'll ever get a bug report that the setting has no effect on brush :laughing:

Could you please add the new option to the example `config.toml` and run `make update-readme`? And perhaps add a remark to the Brush tool section that it'll always use round caps?

I'm also wondering if there's better icons for this, re-using the circle isn't super ideal. But looking at the fluentui icons I don't have a good idea, either.

imizgun:
If somebody will ask about brush not having this effect, I would personally deal with it :)
I've updated `config.toml` and `README.md`, in `README.md` noticed, that `default-round-caps` is `true` for `brush` always: 
```
# Round caps for `arrow` and `line` tools (always true for `brush` tool)
default-round-caps = true
```

Maybe some other ideas on possible icons in toolbar?

robertwidfen:
Tested and looks good to me.

Sorry for accidentally mentioning it in other ticket 🙃

robertwidfen:
Regarding the icons - what do you (two) think about:

image

image

[icons-rounded-caps.patch](https://github.com/user-attachments/files/30596579/icons-rounded-caps.patch)

imizgun:
Both look alright for me

robertwidfen:
> Both look alright for me

Then please change it 🙏🏻 - my patch may work - and then we can merge this.

imizgun:
Done, also resolved merge conflicts :)

Issue #593, opened 2026-07-26
IME/fcitx underline and input box too high


robertwidfen:
### Version

[v0.21.1](https://github.com/Satty-org/Satty/releases/tag/v0.21.1)

### Description

See https://github.com/Satty-org/Satty/issues/2#issuecomment-4818536133

<img width="466" height="183" alt="Image" src="https://github.com/user-attachments/assets/5b702c54-2353-4482-9e52-8658be8444c1" />

robertwidfen:
Looking into this I found more issues. 

- cursor/caret is to high
- bounding box to short at bottom
<img width="327" height="440" alt="Image" src="https://github.com/user-attachments/assets/4f7c6564-2124-47e5-aa8d-2a67938ac4c7" />
<img width="378" height="433" alt="Image" src="https://github.com/user-attachments/assets/5e379037-cbe5-448a-9315-8e85228cf0be" />

- pre-edit highlight, underline and choice list to high
<img width="677" height="400" alt="Image" src="https://github.com/user-attachments/assets/af628562-f69b-4604-9b3a-7cb06e9e830f" />

- pre-edit highlight not aligned to chars, but bytes
<img width="871" height="371" alt="Image" src="https://github.com/user-attachments/assets/fd6b1a18-1f4e-4ed3-9722-dd2b197a1364" />

- selection overlapping with bounding box
<img width="342" height="387" alt="Image" src="https://github.com/user-attachments/assets/6211c360-9715-44f2-a614-e09177767532" />

- line-height to small
<img width="493" height="319" alt="Image" src="https://github.com/user-attachments/assets/8e936305-9386-478d-93bc-fbed45410a8b" />

PR #595, opened 2026-07-31
Fix text cursor and selection position, bounding rectangle, preedit highlight and underline

### LINKED ISSUE #575: Removed text keeps box which is ignored during nudge
### Version

54857978d17e8e67fa94f1c07e5d187dd2b311eb

### Description

Repro:
* use text tool to enter text
* delete all
* observe box is still present
* use nudge shortcut
* box does not move until text is entered

I believe we should just delete the box when no text is left.

robertwidfen:
See #593.

- fix: text cursor & selection placement and line height
- fix: text bounding rectangle glitches
- fix: text ime/pre-edit underline and highlight

@lilydjwg would you be so kind to take a look or test it?

Looks now like + a highlight height fix not in the video

https://github.com/user-attachments/assets/3255fe10-8f01-4160-8c40-e4d98c9199a4

### DIFF
--- a/src/tools/text.rs
+++ b/src/tools/text.rs
@@ -270,6 +270,7 @@ impl Drawable for Text {
         } else {
             (font_metrics.height() / canva_scale).abs()
         };
+        line_height *= 1.2; // increase line height a bit
 
         let mut line_layouts: Vec<LineLayout> = Vec::with_capacity(lines.len());
         let mut baseline = self.pos.y;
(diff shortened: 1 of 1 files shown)


RobertMueller2:
lgtm, let's see if we get any further feedback.

lilydjwg:
Yep, it's fixed!

robertwidfen:
> Yep, it's fixed!

Many thanks for testing 🤗.

PR #598, opened 2026-08-02
feat: round fat arrow when round caps is enabled

robertwidfen:
fix: offsets to start/end not at mouse cursor

See also #592.

The red arrow on left is how fat rounded will look like.

bug-output


RobertMueller2 (approved):
lgtm.

PR #600, opened 2026-08-03
refactor: use custom component for tool buttons to allow grouping

### LINKED ISSUE #538: feature: Tool secondary, tertiary modes
### Description

- Originally from #469, see conversation.
- Has some synergy with #463
- May save some work if done post #450 

Pixelate/blur is a nice example. Pixelate adds pixelate and pseudo-pixelate, and overlaps with blur. Potentially, these could all be under the blur tool.

- how to access secondary modes with mouse? Popup on the button?
- how to access secondary modes with keyboard? direct-access shortcut? toggle shortcut? repeat pressing of existing tool shortcut to toggle mode?
- save state for last selected tool?

Other ideas:
- merge line and arrow tool, add double-end arrow mode
- merge rectangle and ellipse, perhaps filled rectangle and filled ellipse as well 
- add different application modes to highlight (freehand, ...)
- use different shapes for the marker tool
- rotate left and rotate right (that is if we had a rotate tool yet 😛 )
- ...

RobertMueller2:
Closes: #538
prereq for #469

This allows grouping tools under one button. e.g. arrow and line, rectangle and circle. Right-click opens popover.

Used FactoryComponent along with FactoryVecDeque because that simplified reference storage on the toolbar. Controllers can't be cloned, that made a regular component quite tricky here, i've tried for quite a while to use a regular Component, this would likely have required more explicit lifetimes.

The downside of this is, ungrouped tools also have to use the new component. I can't exclude for 100% that a mixed use was possible, but the code would have been messier that way.

- [x] Would like to figure out to show both icon and label on the popover menuitems, but too tired now
- [x] Perhaps the new component can still be optimised a tad for the single tool case

I'm not implementing any variants here, that can be done later and needs new issues per variant. There will be some with pixelate, though.

Video slightly outdated, the popver buttons now show both icon and label.

https://github.com/user-attachments/assets/8ed3e6fd-50f6-41f1-ba7a-7b818fda4671

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -79,6 +79,7 @@ fn main() -> Result<(), io::Error> {
             "arrow-counterclockwise-regular",
             "dismiss-regular",
             "checkmark-regular",
+            "caret-down-right-filled",
         ],
     );
 
--- a/src/tools/mod.rs
+++ b/src/tools/mod.rs
@@ -128,6 +128,13 @@ pub trait Tool {
 }
 
 #[derive(Clone, Debug)]
+pub struct GroupableTool {
+    pub tool: Tools,
+    pub icon_name: String,
+    pub tooltip: String, // should be fine as long as there is no runtime shortcut editing
+}
+
+#[derive(Clone, Debug)]
 pub struct InputContext {
     pub im_context: IMMulticontext,
     pub widget: gtk::Widget,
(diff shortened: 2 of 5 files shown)


robertwidfen:
LGTM

PR #603, opened 2026-08-05
fix: keep text outline stable under zoom

### LINKED ISSUE #596: Text outline "explodes" when zooming in
### Version

_No response_

### Description

See issue #573 and PR #574 which added this feature.

Seems to be only a visual artifact when drawing with `stoke_text()` in the zoomed view and does not affect the exported screenshot. 

https://github.com/user-attachments/assets/50202784-e37f-4552-9545-de1848c423a2

I played with the following, but it does not fix it.
```rust
                    paint.set_line_join(LineJoin::Round);
                    paint.set_line_cap(LineCap::Round);
                    paint.set_miter_limit(1.5);
```

Then tried to paint with `fill_text` multiple times with offsets, but this looks ugly.
Then tried to paint with `fill_text` and larger and smaller font, but could not center then well.

@rokokol - do you have any ideas?

rokokol:
Fixes #596.

The Alt text outline stroked each glyph with a miter-joined stroke. When the canvas is scaled (zoom), femtovg renders text through direct path-stroking of the real glyph contour instead of the glyph atlas, and two issues surfaced there:

- the default **miter** join spikes at the sharp corners of glyph outlines, so the outline visually "explodes" the further you zoom in — this is the reported bug;
- femtovg additionally scales the stroke width by its internal `font_scale` (a quantized canvas average scale) *on top of* the transform, so the outline grew disproportionately with zoom and no longer matched the exported image.

Neither is visible in the export because that path uses only a translation (no scale) and goes through the glyph atlas, where the stroke width is taken raw.

**Fix:** stroke the outline with round joins and divide the width back out by that same `font_scale` (mirroring `quantize(avg_scale, 0.1).min(7.0)`). The border stays smooth and keeps a constant thickness relative to the glyphs at every zoom level, matching the export.

Verified by hand: large text, strong zoom, both `Inverted` and `Contrast` outline modes — outline stays smooth and its thickness is stable across zoom and matches the saved PNG.

### DIFF
--- a/src/tools/text.rs
+++ b/src/tools/text.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use femtovg::{Color, FontId, Paint, Path};
+use femtovg::{Color, FontId, LineJoin, Paint, Path};
 use relm4::gtk::glib::GString;
 use relm4::gtk::prelude::IMContextExt;
 use relm4::gtk::{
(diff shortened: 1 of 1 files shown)


RobertMueller2:
Thank you for the fix. Apparently there's a merge conflict here, did you base this of an older commit of yours?

robertwidfen:
I debugged this a bit. The fix (idea) was working until #589 was merged which is upgrading femtovg from 0.25.0 to 0.26.0.

IMHO looks like a issue in femtovg.

Created a issue there - https://github.com/femtovg/femtovg/issues/311 - maybe they can help.

rokokol:
Yes, I forgot to update my local repository before fix. And, looks like it's definitely problem with femtovg, thanks for the issue!

robertwidfen:
I applied your original changes slightly modified in PR #605.

Can you try it?

It keeps the outline mostly stable - but the fix is odd. I do not understand the internals of femtovg. A fix in femtovg would be better. 🙃

robertwidfen:
A fix in femtovg is coming: https://github.com/femtovg/femtovg/pull/312

So I close this.

rokokol:
Sorry for the silence, been busy lately. Glad to see things worked out in the end :)

PR #605, opened 2026-08-08
fix: keep text outline stable under zoom

### LINKED ISSUE #596: Text outline "explodes" when zooming in

robertwidfen:
Fixes #596 - partly.

But it is odd any should not be like this.

The value to switch scaling was manually determined.

See
- PR #603
- https://github.com/femtovg/femtovg/issues/311

### DIFF
--- a/src/tools/text.rs
+++ b/src/tools/text.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use femtovg::{Color, FontId, Paint, Path};
+use femtovg::{Color, FontId, LineJoin, Paint, Path};
 use relm4::gtk::glib::GString;
 use relm4::gtk::prelude::IMContextExt;
 use relm4::gtk::{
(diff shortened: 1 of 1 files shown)


RobertMueller2:
mhm... I suppose this does prevent the explosion, but the result still very much depends on the zoom level.

https://github.com/user-attachments/assets/a5fd4d83-dddf-45d1-af70-f987e184f0ef

Let's see what we get from the femtovg issue.

robertwidfen:
Interesting - for me it was basically gone, but I have different setup.

Anyway a fix in femtovg is coming: https://github.com/femtovg/femtovg/pull/312

So I close this.

PR #606, opened 2026-08-08
feat: add key bindings to set scale

### LINKED ISSUE #338: Feature: zooming shortcuts/usability

robertwidfen:
Fixes #338

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,13 +258,17 @@ notification-thumbnail = "screenshot"
 "Return" = "run-actions-on-enter"
 "<Control>t" = "toggle-toolbars"
 
-"<Control>f" = "scale:0" # fit to window
+# scale is in %, 0 is fit to window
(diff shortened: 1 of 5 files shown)


RobertMueller2:
I tested this and almost hit the merge button this morning. ;)

Then I started to wonder if we should interpret the scale parameter as percentage, so specific zoomed-in levels become available for shortcuts, too. Doesn't mean we have to give them a default binding, just got a hunch that eventually somebody's going to ask. :smile:  What do you think?

(Alternatively, float is possible of course, but we should probably round that to two digits.)

robertwidfen:
That is a great idea! Thanks.

RobertMueller2:
lgtm

Issue #585, opened 2026-07-12
Layout-independent shortcuts


rokokol:
## Context

I'd like to ask about layout-independent keyboard shortcuts, but I know this has a bit of history, so I wanted to check in rather than just re-request the old behavior.

- #508 removed the previous layout-independent key matching, since it broke Ctrl+Y/Ctrl+Z on German keyboards. The conclusion there was that this kind of thing is arguably better handled at the system/toolkit level rather than per-app.
- #555 reworks keybindings entirely into user-defined `"key" = "tool-or-command"` mappings, and explicitly drops automatic layout independence in favor of "bind what you want."
- #450 is the broader umbrella issue for keybinding design.

## Question

Given all that, I'm curious whether there's still room to revisit this - specifically for non-Latin layouts (Cyrillic, etc.), where basically none of the default single-letter shortcuts work unless you switch to a Latin layout first. That's a bit more disruptive than the DE Ctrl+Y case, since it's not one binding but essentially all of them.

I noticed the #508 discussion mentioned GTK4's `ShortcutController`/physical-key matching as a possible "correct" way to do this, as opposed to the logical-key approach that caused the original regression. Would something like that be worth exploring - maybe just for the shipped defaults, while leaving user-defined bindings in `[keybinds]` as layout-dependent, per #555?

Mostly asking to understand whether this is something the maintainers would consider, or if it's been deliberately ruled out already. Happy to help test/research if it's the former.

RobertMueller2:
#555 does allow binding of commands to more than one shortcut, so in principle, while it's probably inconvenient and tedious, it should be possible to duplicate all bindings to the logical non-latin counterpart once the PR lands.

That said, if there is an easy way to say that a shortcut is physical, or that as an option all shortcuts should be considered physical, we can certainly consider that. As long as we can do it programatically and generic, I don't see why not.

I'm not near a computer right now, is this not a problem in other applications?

rokokol:
> I'm not near a computer right now, is this not a problem in other applications?

Usually it's not a problem. For example, Obsidian allows you to choose a key binding in the Latin layout, and it also works with the corresponding Cyrillic keys.

robertwidfen:
Sorry for the inconvenience - a more proper fix in #508 would have been to duplicate the check and split it up unto two phases instead of checking in one `if`, i.e.
1. Check all layout dependent keys first
2. Check all layout independent keys second

The code also **only** does this for bindings with modifiers - the single letter bindings are not layout independent.

[...]

PR #608, opened 2026-08-10
Fix/use unshifted key for binding

robertwidfen:
Builds on #606 as I found it during testing it.


RobertMueller2:
definitely an improvement, thanks. I'm not sure if this will really survive #585 but let's use this for now.

robertwidfen:
It actually is a bug IMHO once I realized <kbd>Shift+Control+1</kbd> was not working for scaling - well then I actually switched to <kbd>Alt+1</kbd> which is much nicer for one hand.

I already had shifted shortcuts in my config, but missed this one. 🙊

PR #616, opened 2026-08-23
fix: clip crop size to the part inside the image

### LINKED ISSUE #337: Crop gets clamped to image size, but uses full width/height
### Version

0.20.0

### Description

Repro:
1. resize satty so there is some spare area on canvas
2. crop from outside of the image, I assume it can only be reproduced with x < image x offset and y < image y offset

Expected:
Saved image should be the intersection between crop area and the image*

Actual:
Crop offset is moved, but width or height remain unchanged

Image

*) I know we may want to save a larger image if there are annotations on the canvas outside of the image, I think we need to discuss that feature first, though.

hdimer:
Fixes #337. Rebased onto `main` after #607.

### The clipping bug

`rect_ensure_in_bounds` moved the rectangle's top-left corner onto the bound and only afterwards tried to shrink the size by however much had been cut off:

```rust
if pos.x < bounds.0.x {
    pos.x = bounds.0.x;
    size.x -= bounds.0.x - pos.x;   // pos.x == bounds.0.x by now, so this is always -= 0
}
```

`pos.x` had already been reassigned, so the subtraction was unconditionally `-= 0.0`. The left/top branches only ever moved `pos` and never touched `size` at all, which is why review here is cheap: there is no old sizing behaviour to preserve, only dead code to replace. Hence the reported symptom, cropping from outside the image corrected the offset but kept the width/height the user had dragged, so the saved image did not match the visible crop rectangle. Same on the y axis.

The two statements are now swapped so the overhang is measured before `pos` is clamped, and the result is floored at `0.0` — a rectangle lying entirely outside the bound would otherwise come out negative (a 100px wide crop starting 300px left of the image gives -200).

### The dimension readout

Second commit, per your request: the label now reports what would be saved rather than the rectangle that was dragged.

Neither of the two places that fed it had access to the image size, so both now send the crop rect as `SketchBoardInput::CropDimensionsUpdate` and `SketchBoard` clips it before emitting `DimensionsUpdate`; it keeps the image bounds from the pixbuf it is initialised with.

The clip itself is a new `math::crop_rect_in_bounds`, which `render_native_resolution` now uses too. That is the point of naming it: the label and the file come out of the same function, so they cannot drift apart. The rounding order is part of that — `rect_round` then clip differs from clip then round at the edges. A crop at x=190.5 with width 9.9 on a 200-wide image saves 9 px; the readout says 9 now, and would have said 10 the other way round.

Both emit sites read the rect from the crop drawable's own `bounds()`, which is exactly what `render_native_resolution` clips, so there is no second derivation to keep in sync. That retires `Crop::get_rectangle`, whose only caller was the old readout.

`crop_rect_in_bounds` also floors the size at zero. Without that, a crop dragged fully past the right edge would read `-50`, since the lower clamps in `rect_ensure_in_bounds` still return negatives. The save already produces `0` there (`-100.0f32 as usize`), so the readout now agrees with it. I have left `rect_ensure_in_bounds` itself alone so `blur.rs`, its only other caller, is untouched — the symmetric one-liner is still on the table if you want it, but it did not need to ride along with this.

### Gestures the readout follows

Resizing a crop was not the only way to change what gets saved, so the emit is not only on the resize path:

- drawing the crop (`CropTool`, as before)
- resizing it by a handle (`PointerTool`, as before)
- **moving it** with the pointer tool — new here. Dragging a committed crop so it hangs off an edge changes the saved size, and on `main` the label did not move. That was harmless while the label meant "the crop rectangle"; it is not harmless now that it means "what will be saved".
- **nudging it** with the arrow keys (`NudgeSelection`) — same reasoning.

Undo/redo and dismissing a crop are *not* covered, and that is worth flagging: `DimensionsUpdate(None)` resolves to the full image dimensions in `main.rs`, but nothing in the tree sends `None` any more. `CropTool::handle_dismissed` used to, and it was removed in #607. So after undoing a crop the save reverts to the full image while the label keeps the crop's numbers. That is a pre-existing gap rather than something this PR introduces, and the affordance to fix it already exists, but I did not want to fold an unrelated regression into this diff. Happy to send it separately.

### What this means for a selection with no overlap

A crop dragged entirely off the left of the image now clips to a zero width instead of keeping its dragged width, so it reaches `create_image_empty(0, h)`.

That path is not new. On current `main`, a crop dragged entirely to the *right* never enters the top/left branches at all, hits `pos.x + size.x > bounds.1.x`, and comes out with `size.x = -100`; `-100.0f32 as usize` saturates to `0`, so the same call already happens today. This change adds the left/top route to it, a clean `0` rather than a wrong-but-positive width. It fails politely either way, `render_native_resolution`'s `Err` is handled with an `eprintln!` and a `Propagation::Stop`, not a panic.

Preventing the save outright is left to the separate PR you mentioned. What the readout change adds in the meantime is that the user can now see the `0` before they try.

### Second call site

`Blur::draw` uses `rect_ensure_in_bounds`, so a blur box dragged in from outside the image now clips instead of keeping its full width, and inherits the #337 fix for free. The committed path early-returns on a non-positive size; the editing preview does not guard, and draws a `rounded_rect` with a non-positive size. Flagging it since the issue only talks about crop.

### Tests

`src/math.rs` had no test module, so this adds one, following the pattern already in `sketch_board.rs`.

Five for `rect_ensure_in_bounds`: clipping on both axes at once, the two fully-outside cases, a rect straddling all four edges, and a fully-inside non-regression case. The first three fail on current `main`; the last two pass either way and are there to pin the function against a future refactor.

Four for `crop_rect_in_bounds`: the clip, the rounding order, and one per axis of the zero floor.

Note the lint workflow runs fmt, clippy and doc but not `cargo test`, so these are compiled and linted on CI but not executed.

### GenAI

Written with AI assistance, disclosed per CONTRIBUTING. I understand and vouch for it, it's a statement reorder, a lower bound, and a message hop to the one component that knows the image size.

I ran the tests against unpatched `main` to confirm they fail there, and checked each assertion earns its place by re-breaking the fix one piece at a time. For `rect_ensure_in_bounds`, dropping either `.max(0.0)` breaks exactly one test and reverting the y-axis reorder breaks two. For `crop_rect_in_bounds`, dropping `rect_round` breaks only the rounding test, dropping either half of the zero floor breaks only that axis's test, and dropping the clip breaks three. The per-axis floor tests exist because a first pass only covered x, and a mutant that clamped x but not y passed the whole suite.

One caveat worth stating plainly: I'm on macOS and can't build the GTK side of the crate. `src/math.rs` is pure std, so those tests really run. Everything outside it is `cargo fmt --check` clean against the repo, and clippy-clean only as isolated expressions in a scratch crate carrying the real signatures. CI's clippy job is the real check on the wiring, so it's worth letting it run before you read closely.

---
Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -23,7 +23,7 @@ use resource::resource;
 use crate::{
     APP_CONFIG,
     configuration::Action,
-    math::{Vec2D, rect_ensure_in_bounds, rect_round},
+    math::{Vec2D, crop_rect_in_bounds},
     sketch_board::SketchBoardInput,
     tools::{Drawable, RenderingMode, Tool, Tools},
 };
@@ -520,10 +520,8 @@ impl FemtoVgAreaMut {
             .iter()
             .find(|d| d.get_rendering_mode() == RenderingMode::Crop)
             .and_then(|d| {
-                d.bounds().map(|(tl, br)| {
-                    let rect = (tl, br - tl);
-                    rect_ensure_in_bounds(rect_round(rect), bounds)
-                })
+                d.bounds()
+                    .map(|(tl, br)| crop_rect_in_bounds((tl, br - tl), bounds))
             })
             .unwrap_or(bounds);
 
--- a/src/math.rs
+++ b/src/math.rs
@@ -268,14 +268,16 @@ pub fn rect_ensure_positive_size(pos: Vec2D, size: Vec2D) -> (Vec2D, Vec2D) {
 pub fn rect_ensure_in_bounds(rect: (Vec2D, Vec2D), bounds: (Vec2D, Vec2D)) -> (Vec2D, Vec2D) {
     let (mut pos, mut size) = rect;
 
+    // The part sticking out has to be measured before pos is moved onto the
+    // bound, otherwise the difference is always zero and size stays untouched.
     if pos.x < bounds.0.x {
+        size.x = (size.x - (bounds.0.x - pos.x)).max(0.0);
         pos.x = bounds.0.x;
-        size.x -= bounds.0.x - pos.x;
     }
 
     if pos.y < bounds.0.y {
+        size.y = (size.y - (bounds.0.y - pos.y)).max(0.0);
         pos.y = bounds.0.y;
-        size.y -= bounds.0.y - pos.y;
     }
 
     if pos.x + size.x > bounds.1.x {
@@ -297,4 +299,51 @@ pub fn ensure_bounding_box(a: Vec2D, b: Vec2D) -> (Vec2D, Vec2D) {
 pub fn rect_round(rect: (Vec2D, Vec2D)) -> (Vec2D, Vec2D) {
     let (pos, size) = rect;
     (pos.round(), size.round())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{Vec2D, rect_ensure_in_bounds};
+
+    fn bounds() -> (Vec2D, Vec2D) {
+        (Vec2D::zero(), Vec2D::new(200.0, 100.0))
+    }
+
+    #[test]
+    fn rect_ensure_in_bounds_clips_both_axes_when_overlapping_top_left() {
+        let rect = (Vec2D::new(-40.0, -30.0), Vec2D::new(100.0, 80.0));
+        assert_eq!(
+            rect_ensure_in_bounds(rect, bounds()),
+            (Vec2D::zero(), Vec2D::new(60.0, 50.0))
+        );
+    }
+
+    #[test]
+    fn rect_ensure_in_bounds_removes_width_when_rect_is_left_of_bounds() {
+        let rect = (Vec2D::new(-300.0, 10.0), Vec2D::new(100.0, 20.0));
+        let (_, size) = rect_ensure_in_bounds(rect, bounds());
+        assert_eq!(size.x, 0.0);
+    }
+
+    #[test]
+    fn rect_ensure_in_bounds_removes_height_when_rect_is_above_bounds() {
+        let rect = (Vec2D::new(10.0, -300.0), Vec2D::new(20.0, 100.0));
+        let (_, size) = rect_ensure_in_bounds(rect, bounds());
+        assert_eq!(size.y, 0.0);
+    }
+
+    #[test]
+    fn rect_ensure_in_bounds_clips_rect_straddling_all_four_edges() {
+        let rect = (Vec2D::new(-50.0, -50.0), Vec2D::new(400.0, 300.0));
+        assert_eq!(
+            rect_ensure_in_bounds(rect, bounds()),
+            (Vec2D::zero(), Vec2D::new(200.0, 100.0))
+        );
+    }
+
+    #[test]
+    fn rect_ensure_in_bounds_leaves_rect_inside_bounds_unchanged() {
+        let rect = (Vec2D::new(10.0, 20.0), Vec2D::new(30.0, 40.0));
+        assert_eq!(rect_ensure_in_bounds(rect, bounds()), rect);
+    }
 }
--- a/src/math.rs
+++ b/src/math.rs
@@ -296,6 +296,13 @@ pub fn ensure_bounding_box(a: Vec2D, b: Vec2D) -> (Vec2D, Vec2D) {
     (a.min(b), a.max(b))
 }
 
+// The part of a crop rectangle that is actually inside the image: what a save
+// writes out, and therefore what the dimension readout has to be derived from.
+pub fn crop_rect_in_bounds(rect: (Vec2D, Vec2D), bounds: (Vec2D, Vec2D)) -> (Vec2D, Vec2D) {
+    let (pos, size) = rect_ensure_in_bounds(rect_round(rect), bounds);
+    (pos, size.max(Vec2D::zero()))
+}
+
 pub fn rect_round(rect: (Vec2D, Vec2D)) -> (Vec2D, Vec2D) {
     let (pos, size) = rect;
     (pos.round(), size.round())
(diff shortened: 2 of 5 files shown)


RobertMueller2:
Thanks for looking into this and fixing the issue.

Just a heads up, I anticipate some overlap with #607. Which is a bigger feature that has taken quite a bit of work and is almost at the finish line. We'll probably merge #607 first.

hdimer:
No problem, #607 first makes sense.

Both hunks here are small and independent of that work, so this should rebase cleanly whenever it lands. Two things I noticed diffing the branches, in case they're useful for #607 itself:

- `rect_ensure_in_bounds` is unchanged in #607, so the ordering slip behind #337 survives it. The new call site (`rect_ensure_in_bounds(rect_round(rect), bounds)`) goes through the same code, so a crop started left of the image still keeps its dragged width instead of clipping to the intersection.
- #607 drops the `.filter(|(_, size)| !size.is_zero())` in `render_native_resolution`, and the new `Crop::bounds()` always returns `Some`, so nothing guards `create_image_empty` against a collapsed axis in the new path.

Happy to rebase this once #607 is merged, or to close it and let you fold the two-line math fix into #607 if that's less work for you. Whichever you prefer.

RobertMueller2:
With #607 merged, I think we can merge this one after a rebase.

Regarding the dropped check in `render_native_resolution`, I'm not 100% sure. Right now, if there's no valid selection, we get the full image, I feel that's kinda unexpected too. Perhaps if we don't have a valid selection, we should prevent saving altogether?

hdimer:
Rebased. The math fix applies unchanged, the three clipping tests still fail on `main` and pass here, so #607 didn't touch that path. I've also added a straddling-all-four-edges case and renamed the tests to match the style in `sketch_board.rs`.

I dropped the second hunk. #607 removed the `is_zero()` filter outright, so re-adding a per-axis version would be me answering your question rather than fixing #337. This PR is now just `rect_ensure_in_bounds` plus its tests.

On the question itself: I'd go with preventing the save. Handing back the full image when the user has dragged a selection containing none of it is the one result they definitely didn't ask for.

Two things worth knowing before you decide, both of which I've written into the PR body as well:

The degenerate render target is already reachable on `main` without this PR. A crop dragged entirely to the *right* of the image never enters the top/left branches at all, it hits `pos.x + size.x > bounds.1.x` and comes out with `size.x = -100`, and `-100.0f32 as usize` is 0. So `create_image_empty(0, h)` happens today. What this PR adds is the left/top route to the same place. So wherever the check lands, it probably wants to sit on the final `(pos, size)` rather than on the crop drawable, since the negative sizes come out of those two lower clamps.

Separately, the dimension readout has the same blind spot: `emit_crop_dimensions_update` reports `get_rectangle()` and the resize preview reports the raw `new_br - new_tl`, neither clipped to the image, so the label can disagree with what gets saved. Also pre-existing on the right/bottom side, but it sits in the same "what is a valid selection" space you're weighing.

One other thing, take it or leave it: the whole function is a rect intersection and collapses to

```rust
let top_left = pos.max(bounds.0);
let bottom_right = (pos + size).min(bounds.1);
(top_left, (bottom_right - top_left).max(Vec2D::zero()))
```

which removes the ordering hazard structurally instead of commenting on it, and makes the lower clamps non-negative too. I've kept it out of this PR because it's a rewrite of a function you've already said you're happy to merge, and it touches the behaviour you're still deciding on. Say the word and I'll either fold it in here or send it separately.

RobertMueller2:
OK, I think we should prevent saving, happy to leave that for another PR.

And I agree, the readout should reflect what would be saved, not the dimensions of the crop rectangle. Can you add that here please?

hdimer:
Added, pushed as a second commit.

Neither of the two places feeding the label knew the image size, so both now send the crop rect as `SketchBoardInput::CropDimensionsUpdate` and `SketchBoard` clips it before emitting `DimensionsUpdate` — it holds the image bounds from the pixbuf it's initialised with. The clip is a new `math::crop_rect_in_bounds` that `render_native_resolution` uses too, so the label and the file come out of one function rather than two that have to agree. The rounding order turned out to matter: a crop at x=190.5 with width 9.9 on a 200-wide image saves 9, and clip-then-round would have printed 10.

Both sites now read the rect from the crop drawable's `bounds()`, the same one the render clips, which also retired `Crop::get_rectangle` — the old readout was its only caller.

Three things you should know before you read it:

**It emits on move and nudge too, not just resize.** A committed crop is selectable, so you can drag it or arrow-key it off an edge, and that changes what gets saved. That was harmless while the label meant "the crop rectangle"; once it means "what will be saved" it's the same bug you reported, via a different gesture. Six lines in the same style, but it's scope I added rather than scope you asked for, so say the word if you'd rather I split it out.

**`crop_rect_in_bounds` floors the size at zero**, otherwise a crop dragged fully past the right edge reads `-50`, since the lower clamps still return negatives. The save already gives 0 there. I deliberately did *not* fix that asymmetry inside `rect_ensure_in_bounds` — you've already said you're happy to merge that function, and touching it would drag `blur.rs` along. The symmetric one-liner is still yours for the asking, just not welded to this.

**Undo and dismiss are still not covered, and that's a pre-existing regression worth knowing about.** `DimensionsUpdate(None)` resolves to the full image dimensions in `main.rs`, but nothing sends `None` any more — `CropTool::handle_dismissed` did, and it went away in #607. So undoing a crop reverts the save to the full image while the label keeps the crop's numbers. Not this PR's doing and I didn't want to fold it in, but it's the same invariant, and the plumbing for it already exists. Separate PR if you want it.

Tests: four more in `src/math.rs`, one per piece of the new function. Two of them are the per-axis zero floor, because my first pass only asserted x and a mutant that clamped x but not y passed the entire suite.

One caveat: I'm on macOS and can't build the GTK side, so only the `math.rs` tests actually execute here. The rest is `cargo fmt --check` clean against the repo and clippy-clean only as isolated expressions in a scratch crate with the real signatures. Worth letting CI run before you read closely.

RobertMueller2:
> **Undo and dismiss are still not covered, and that's a pre-existing regression worth knowing about.** `DimensionsUpdate(None)` resolves to the full image dimensions in `main.rs`, but nothing sends `None` any more — `CropTool::handle_dismissed` did, and it went away in #607. So undoing a crop reverts the save to the full image while the label keeps the crop's numbers. Not this PR's doing and I didn't want to fold it in, but it's the same invariant, and the plumbing for it already exists. Separate PR if you want it.

Thanks for pointing that out, we'll open an issue for that for now.

RobertMueller2:
lgtm, thanks

PR #624, opened 2026-09-01
fix: scale:100 button on toolbar

### LINKED ISSUE #623: scale:100 works as a shortcut, but not from toolbar
### Version

85b5742  but probably way earlier

### Description

This is off by 100 in `sketch_board.rs`. If I click the button, this happens. Shortcut on the other hand works fine.

Image

RobertMueller2:
closes: #623


robertwidfen:
Actually I missed it in #606 🙃

PR #625, opened 2026-09-01
fix(build): replace epoxy-rs

### LINKED ISSUE #615: "cargo install satty" fails: yanked xml-rs versions via gl_generator/epoxy crates
### Version

trying to install satty v0.22.0 via cargo

### Description

```
cargo install satty
```
fails to resolve dependencies:
```
error: failed to compile `satty v0.22.0`, intermediate artifacts can be found at `/tmp/cargo-installh6XGJQ`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_BUILD_BUILD_DIR` to that path.

Caused by:
  failed to select a version for the requirement `xml-rs = "^0.7.0"`
    version 0.7.0 is yanked
    version 0.7.1 is yanked
  location searched: crates.io index
  required by package `gl_generator v0.9.0`
      ... which satisfies dependency `gl_generator = "^0.9"` of package `epoxy v0.1.0`
      ... which satisfies dependency `epoxy = "^0.1.0"` of package `satty v0.22.0`
```

#### Workaround 
build from git using the repo's own lockfile (must use --locked, or cargo re-resolves and hits the yank again):
```
git clone https://github.com/Satty-org/Satty
cd Satty
cargo install --path ./ --locked
```

Note: this only works as long as the committed Cargo.lock stays valid against Cargo.toml.

RobertMueller2:
Closes: #615

epoxy-rs is 8 years old. One of its transient dependencies (xml-rs) was pulled recently, making it impossible to build Satty via cargo install.

A fix is available upstream at https://github.com/mjkoo/epoxy-rs/pull/10, but we don't know if it'll get merged. We could have created a vendor fork from the above epoxy-rs PR, however we're really only using `get_proc_addr`, which provides symbol names for libepoxy.

The code is very similar between epoxy-rs and gtk4-rs examples. Moreover, the relevant gtk4-rs example was also replaced with direct use of libepoxy, see https://github.com/gtk-rs/gtk4-rs/commit/bf5186fe730407db387fa09482a6f2c86981abe8

This PR adopts some of the example code into Satty and removes the need to use epoxy-rs.

The code could be a bit nicer if we used `OnceLock::get_or_try_init`, however that is a nightly feature. -> #626

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,12 +93,6 @@ checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
-
-[[package]]
-name = "bitflags"
 version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b588b76d00fde79687d7646a9b5bdf3cc0f655e0bbd080335a95d7e96f3587da"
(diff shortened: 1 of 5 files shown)


robertwidfen:
lgtm - thanks

PR #631, opened 2026-09-05
Fix editing of crop and text

robertwidfen:
Fixes some issues from PR #607 - see issue #159. 

- feat: when switching to crop tool select active crop
- fix: double draw of crop while creating
- fix: preserve text properties when editing by pointer tool


RobertMueller2 (approved):
Makes sense and lgtm.

Thanks!

PR #635, opened 2026-09-06
fix: sleep for a bit if there was a recent notifcation thumbnail

### LINKED ISSUE #634: [Bug] `--notification-thumbnail screenshot` shows generic fallback icon due to `TEMP_DIR` deletion on exit
### Version

0.22.0 (and current main branch)

### Summary
Hi! First of all, thank you very much for Satty and for your work :)

When using `--notification-thumbnail screenshot` (especially combined with `--early-exit`), the desktop notification daemon displays a generic application icon (`com.gabm.satty` or  `dialog-information`) instead of the screenshot preview thumbnail.

---

### Steps to Reproduce
1. Run Satty with early exit and screenshot thumbnail enabled:      
 `satty --filename /path/to/image.png --early-exit --notification-thumbnail screenshot --copy-command wl-copy
 `
3. Trigger the copy or save action (e.g. press Enter or click the copy button).
4. Observe the notification delivered to the desktop notification daemon (e.g. SwayNC, Quickshell, Mako, Dunst).
---
### Expected Behavior

The notification should display the scaled thumbnail preview of the captured screenshot.

---
### Actual Behavior

The notification daemon displays a generic fallback icon (such as the app icon or dialog-information).

---
### Root Cause Analysis

1. How the thumbnail is sent:
  In src/notification.rs (log_result_with_pixbuf), Satty scales down the pixbuf and writes it as a temporary PNG file using tempfile::Builder::new().disable_cleanup(true).tempfile_in(d.
  path()), where d is the global TEMP_DIR (tempfile::TempDir).
  The file is attached as a gio::FileIcon and sent via gio::Notification, which passes image-path = "/path/to/thumbnail.png" in the D-Bus Notify call hints.
2. Why it fails on exit:
  In src/main.rs (lines 572–585):

```
    app.run::<App>(image);

    match TEMP_DIR.write() {
        Ok(mut temp_dir) => {
            if let Some(dir) = temp_dir.take()
                && let Err(e) = dir.close()
            {
                eprintln!("Failed to close temporary directory: {}", e);
            }
        }
        ...
    }

```

When **--early-exit** is **enabled** (or as soon as the user finishes and closes the window), app.run returns immediately and dir.close() removes the entire temporary directory.Because Satty exits within milliseconds after issuing the notification, the temporary thumbnail file is deleted from disk before the asynchronous notification daemon can open and render it.

Consequently, the notification daemon encounters a missing file and falls back to the generic application icon.

---
### Proposed Solution

Instead of creating the temporary notification thumbnail inside the ephemeral **TEMP_DIR** that gets wiped on application shutdown, create it directly in the system temporary directory (std::env::temp_dir()) with disable_cleanup(true).

This allows the thumbnail file to persist in **/tmp** long enough for the notification daemon to read and display it, while leaving standard temporary file cleanup to the OS.

#### Diff / Patch:

    diff --git a/src/notification.rs b/src/notification.rs
    index b52ab48..4fc8eb9 100644
    --- a/src/notification.rs
    +++ b/src/notification.rs
    @@ -1,7 +1,5 @@
     use relm4::gtk::gio::{FileIcon, Icon};
     use relm4::gtk::gio::{Notification, prelude::ApplicationExt};
    -
    -use crate::TEMP_DIR;
     use crate::configuration::APP_CONFIG;
     use relm4::gtk::gdk_pixbuf::{InterpType, Pixbuf};
     use relm4::gtk::prelude::Cast;
    @@ -47,26 +45,20 @@ pub fn log_result_with_pixbuf(msg: &str, pixbuf: Pixbuf) {
         // we can't just use a tempfile here because we need the path for the FileIcon.
         // Also, we can't just use NamedTempFile with cleanup, because it gets dropped
         // at the end of this function, which can be too early for the notification daemon.
    -    let tempfile: Option<NamedTempFile> = match TEMP_DIR.read() {
    -        Ok(guard) => match &*guard {
    -            Some(d) if pixbuf.is_some() => {
    -                if let Ok(tf) = tempfile::Builder::new()
    -                    .disable_cleanup(true)
    -                    .suffix(".png")
    -                    .tempfile_in(d.path())
    -                {
    -                    Some(tf)
    -                } else {
    -                    eprintln!("Could not create temporary file");
    -                    None
    -                }
    -            }
    -            _ => None,
    -        },
    -        Err(e) => {
    -            eprintln!("Error acquiring read guard for temp directory: {}", e);
    +    let tempfile: Option<NamedTempFile> = if pixbuf.is_some() {
    +        if let Ok(tf) = tempfile::Builder::new()
    +            .prefix("satty-thumb-")
    +            .suffix(".png")
    +            .disable_cleanup(true)
    +            .tempfile()
    +        {
    +            Some(tf)
    +        } else {
    +            eprintln!("Could not create temporary file");
                 None
             }
    +    } else {
    +        None
         };

         let icon = match tempfile.as_ref() {

I have to say that all this analysis has been done via Gemini because thumbnails aren't working currently on my system. I have tested the solution on NixOS / Hyprland with Quickshell via an overlay which patches the code dynamically; with this patch, thumbnails render immediately and reliably.

Thank you very much for your effort and for your time!

RobertMueller2:
Closes: #634

Notification thumbnails are files referenced in dbus notifications by path. Since these are temporary files, we clean them up. However, when using early exit upon save, this might be before the notification daemon can read them. Unfortunately, there is no feedback mechanism.

With this PR, satty reads the latest file's mtime in the TempDir and determines whether that file was created long ago.

The value of 250ms seems rather long on my systems, I never notice any issue with single digit ms values even without this patch. But on slower systems, it might not be enough.

In worst case, this needs a configurable value, or at least useful documentation.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -83,6 +83,9 @@ app-id = "org.satty.satty"
 # experimental feature (0.22.0): show thumbnail as notifcation icon
 # notification-thumbnail = "app-icon"
 notification-thumbnail = "screenshot"
+# how long we are waiting after notification thumbnail was sent before exiting (exit early or quit). This is based on the temporary file mtime.
+# for fast machines, the default value of 250ms is ridiculously high.
+notification-grace-period = 250
 
 # Generic keyboard shortcuts (NEXTRELEASE)
 [keybinds]
--- a/src/configuration.rs
+++ b/src/configuration.rs
@@ -1,13 +1,13 @@
+use clap::Parser;
+use hex_color::HexColor;
+use relm4::SharedState;
+use std::time::Duration;
 use std::{
     collections::HashMap,
     fs,
     io::{self, Write},
     path::Path,
 };
-
-use clap::Parser;
-use hex_color::HexColor;
-use relm4::SharedState;
 
 use serde::Deserialize;
 use serde::de::Deserializer;
@@ -76,6 +76,7 @@ pub struct Configuration {
     title: Option<String>,
     app_id: Option<String>,
     notification_thumbnail: NotificationThumbnail,
+    notification_grace_period: Duration,
 }
 
 #[derive(Default)]
@@ -362,6 +363,9 @@ impl Configuration {
         }
         if let Some(v) = general.notification_thumbnail {
             self.notification_thumbnail = v;
+        }
+        if let Some(v) = general.notification_grace_period {
+            self.notification_grace_period = std::time::Duration::from_millis(v);
         }
 
         // --- deprecated options ---
@@ -672,6 +676,10 @@ impl Configuration {
     pub fn notification_thumbnail(&self) -> NotificationThumbnail {
         self.notification_thumbnail
     }
+
+    pub fn notification_grace_period(&self) -> Duration {
+        self.notification_grace_period
+    }
 }
 
 impl Default for Configuration {
@@ -714,6 +722,7 @@ impl Default for Configuration {
             title: None,
             app_id: None,
             notification_thumbnail: NotificationThumbnail::default(),
+            notification_grace_period: Duration::from_millis(250),
         }
     }
 }
@@ -785,6 +794,7 @@ struct ConfigurationFileGeneral {
     title: Option<String>,
     app_id: Option<String>,
     notification_thumbnail: Option<NotificationThumbnail>,
+    notification_grace_period: Option<u64>,
 
     // --- deprecated options ---
     right_click_copy: Option<bool>,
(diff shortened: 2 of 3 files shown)


robertwidfen:
lgtm

Odd that this is not fixed in the API - always confuses me that we need to
still keep it after the notification was sent 🙃 ...

RobertMueller2:
> lgtm
> 
> Odd that this is not fixed in the API - always confuses me that we need to
> still keep it after the notification was sent 🙃 ...

And for having an external dependency we have alot of own code around this. I'm not super happy with GNotification.

PR #636, opened 2026-09-06
fix: remove leftover icons.toml

### LINKED ISSUE #483: leftover icons.toml
https://github.com/Satty-org/Satty/commit/8e5e559c62f0912d2163995cff54de9481b4bb68 updated relm4-icons and added relm4-icons-build. This includes the icon names in build.rs. Apparently I forgot to remove icons.toml while doing this.

RobertMueller2:
Closes: #483

The source of truth is in build.rs, and that's the only thing currently used by Relm4.

If this changes down the road, we should generate one from the other, not maintain two files.


robertwidfen:
🥳 lgtm

PR #640, opened 2026-09-07
fix: prevent zero-sized or out-of-bounds crop selection

### LINKED ISSUE #637: Crop outside image bounds shows wrong dimension result and emits blank image
### Version

86d413e2

### Description

See also #616 

Image

robertwidfen:
Closes #637

- Return zero rect in math helper if crop is completely outside bounds.
- Return error in render_native_resolution if crop size evaluates to zero.


RobertMueller2 (approved):
lgtm

PR #563, opened 2026-07-02
feat: image tool — insert and paste images into the screenshot

rokokol:
Closes #233

This adds an **Image** tool that inserts pictures into the screenshot as annotations. Once inserted, an image is moved, resized and layered by the pointer tool like any other annotation, through the editing support from #607

https://github.com/user-attachments/assets/666a3345-7684-4d6b-ba34-399d54f8ee8d

The recording predates #607: the tool no longer has handles of its own, and rotation is left for a separate change covering all tools

## Usage

- Click the screenshot with the image tool (toolbar button or <kbd>o</kbd>) to open a file chooser; the picked image is centred on the click and scaled down to at most half of the screenshot, small images are not enlarged. Near the edge it is moved just enough to stay whole on the screenshot
- Drag a box instead of clicking, and the image is fitted into it with its aspect ratio kept; <kbd>Alt</kbd> and <kbd>Shift</kbd> shape the box as they do for rectangle and crop
- <kbd>Ctrl+V</kbd> pastes an image or a copied image file from the clipboard, centred on the pointer, without switching tools. The text tool keeps its own <kbd>Ctrl+V</kbd> while it has input focus
- Dropping an image or an image file onto the window inserts it where it was dropped
- Formats are those gdk-pixbuf has a loader for, plus webp

## Configuration

- New default keybinds: `"o" = "image"` and `"<Control>v" = "paste-image"`, both rebindable through the keybinding registry from #555
- `image` is accepted by `--initial-tool` / `initial-tool`
- README and `config.toml` updated, new entries marked `NEXTRELEASE`

## Implementation notes

- `Image` implements `bounds`, `hit_test`, `translate` and `resize_bounds`, so the pointer tool handles it; the image tool itself only picks the file and where it goes
- Where it goes is an `ImagePlacement`, `Center` or `Fit`, in image coordinates. The tool sends it through `SketchBoardInput::ImagePlaced`, while paste and drop send `SketchBoardInput::ImageSelected` with a canvas position, which the sketch board converts. A centred image is kept whole on the screenshot by its layout, a box is left where it was drawn
- A click arrives on press, before it is known whether the press becomes a drag, so the tool acts on the release of the drag gesture that every press also starts
- `image` is not a new dependency: femtovg's colour emoji already pull it in, and this only enables its `webp` feature, for formats gdk-pixbuf has no loader for
- The renderer-private background upload became a reusable `create_image_from_pixbuf`
[...]

PR #645, opened 2026-09-08
feat: configurable interpolation mode for background image

### LINKED ISSUE #613: Pixel-perfect image rendering and precise crop measurement?
### Description

Hello!

First of all, thank you for this great tool! I use it very regularly for my screenshots, and it has become a really handy part of my workflow.

I'd like to suggest two small features that would make Satty even more useful for inspecting and measuring small areas of screenshots.

1. Enable/disable image anti-aliasing

It would be really useful to have a button or option to toggle image anti-aliasing (more precisely, image interpolation) on and off. Ideally, this option would also have a configurable keybind, so it could be toggled quickly while inspecting an image.

The main use case is inspecting individual pixels or very small areas at high zoom. This should preferably affect only the screenshot/image itself, not the anti-aliasing of the drawing/annotation tools.

2. Precise measurement of very small selections

It would also be great to have a dedicated measurement tool. Alternatively, the existing Crop tool could be improved so that very small selections can be measured reliably. The Crop tool already displays the dimensions of the current selection; however, its 2 px border is drawn centered on the selection boundary. As a result, part of the border overlaps the selected area.

This becomes problematic for very small selections. For example, with a 1×2 px selection, the border can completely obscure the pixels that are supposed to be measured.

A possible solution would be to:

* draw the crop border entirely outside the selected rectangle;
* use a 1 px, non-antialiased border;
* or otherwise make the selection boundary visually distinct without covering pixels inside the selection.

The important point is that a selection such as 1×2 px or 2×2 px should remain completely visible while its dimensions are displayed. Maybe a dedicated measurement tool would also be a great solution...

Thanks again for this great tool, and have a nice day!

RobertMueller2:
Closes: #613

The first commit just added config/cli for a static image interpolation determined per start. In the spirit of #613 I believe a toggle may be more interesting/important.

But while the first variant was simple, this created a new challenge of actually replacing the image ids. Rather than repeatedly dealing with create/update/delete etc we're now keeping two images. It's a bit wasteful from a memory point of view, but in theory this skips some code in `render_background_image` on repeated try and should be a tad faster. I'm not sure if there is a noticeable difference. If anyone's bored, try the other and compare the performance difference ;)

I did not add a toolbar icon but a shortcut (ctrl+i).

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,7 @@ interpolation = "linear" #default
 "s" = "focus-annotation-size-factor"
 "f" = "toggle-fill"
 "k" = "toggle-round-caps"
+"<Ctrl>+i" = "toggle-interpolation"
 
 # Font to use for text annotations
 [font]
--- a/README.md
+++ b/README.md
@@ -257,6 +257,12 @@ app-id = "org.satty.satty"
 # experimental feature (0.22.0): show thumbnail as notifcation icon
 # notification-thumbnail = "app-icon"
 notification-thumbnail = "screenshot"
+# how long we are waiting after notification thumbnail was sent before exiting (exit early or quit). This is based on the temporary file mtime.
+# for fast machines, the default value of 250ms is ridiculously high.
+notification-grace-period = 250
+# set interpolation for the background image
+# interpolation = "nearest-neighbor"
+interpolation = "linear" #default
 
 # Generic keyboard shortcuts (NEXTRELEASE)
 [keybinds]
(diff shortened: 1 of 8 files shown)


RobertMueller2:
just realised this needs rebasing after #563 is being merged. :D

And with that I might need to reconsider this lazy approach here. -> Draft for now

RobertMueller2:
superseded by #652

PR #646, opened 2026-09-08
fix: always pixelate and fringe by dynamic block size

robertwidfen:
See #469.

The blocksize is adjusted a bit and remainder pixels extend blocks at the end of each line resp. row.

Also blocksize is croped to image width and height.

This way pixelate can always be drawn.

Also update the config.toml and README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,10 @@ Highlight:
 - Hold <kbd>Ctrl</kbd> to switch between block and freehand mode (default configurable, see below).
 - Hold <kbd>Shift</kbd> in freehand mode for a straight 15° aligned line. Stop at some position and release and hold <kbd>Shift</kbd> again to achieve perfectly aligned turns.
 
-Additional note on fringe inpaint, pixelate and fringe inpaint + pixelate <sup>NEXTRELEASE</sup>:
+Additional note on blur, fringe inpaint, pixelate and fringe inpaint-pixelate <sup>NEXTRELEASE</sup>:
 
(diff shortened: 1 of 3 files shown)


robertwidfen:
Existing behavior 

https://github.com/user-attachments/assets/bee3c9ea-7e40-4925-9f3c-13d240504e91

New with this PR.

https://github.com/user-attachments/assets/9f6d6c2e-6c5f-439b-ae33-7835434f6d45

RobertMueller2 (approved):
I still don't like that the actual block size varies so much, that was also my reason to not include your suggestion in the #607 PR.  But once #648 lands, satty will offer more assistance achieving predictable results, and with this whole change, it's just a lot easier to use, I think that wins by far over my concerns. Especially the remainder part was not very intuitive. 

Let's get this merged when you're ready.

robertwidfen:
I feel your your dislike - had it too until testing other tools: 😉
- Greenshot keeps the size stable but uses half block size on top and left edges and remainder on right and bottom - not sure how or if it ensures enough pixels per block. If the image is large enough one can steal them from other inner pixels.
- Flameshot/Ksnip/Tensaku all use a dynamic block size.

So dynamic blocksize seem to be the standard as the result looks more uniform than a approach with constant block size. 

And yes my main concern was easy usage. I still kept the is_renderable() stuff intentionally even if it does nothing anymore, but it can be useful for the future and other tools.

In comparison to the other tools using dynamic block size - I try to reduce jitter by stretching only blocks towards the right / bottom to compensate for the remainder - but this is only a subtle difference when main blocksize does not require a change ...

https://github.com/user-attachments/assets/32ef9cee-164c-41fb-b391-e68bfecb7c42

PR #647, opened 2026-09-08
feat!: add predefined aspect ratios to Shift modifier

### LINKED ISSUE #633: Crop to specific aspect ratio/dimensions
### Description

It would be nice if the cropping tool had presets for aspect ratio lock (e.g. 16:9, 4:3) and/or if an aspect ratio (or image dimensions) to lock the crop to could be manually set.

robertwidfen:
Closes #633.

Show image resp. tool aspect ratio in toolbar.
Keep mouse cursor on shape where possible.

Also refactors dimension updates for toolbar - passing an unused value makes no sense.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,10 @@
 {
-  "rust-analyzer.linkedProjects": ["./Cargo.toml"],
-  "rust-analyzer.checkOnSave.command": "clippy"
+  "rust-analyzer.linkedProjects": [
+    "./Cargo.toml"
+  ],
+  "rust-analyzer.checkOnSave.command": "clippy",
+  "[toml]": {
+    "editor.tabSize": 4,
+    "editor.insertSpaces": true
+  }
 }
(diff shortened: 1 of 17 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,10 @@
 {
-  "rust-analyzer.linkedProjects": ["./Cargo.toml"],
-  "rust-analyzer.checkOnSave.command": "clippy"
-}
+  "rust-analyzer.linkedProjects": [
+    "./Cargo.toml"
+  ],
+  "rust-analyzer.checkOnSave.command": "clippy",
+  "[toml]": {
```
RobertMueller2: I'm not using vscode very often here, but probably a good idea. 😄 

I'm not sure this was intended, though?

RobertMueller2:
if I'm using shift, I can not draw into all 4 directions. E.g. I start with a rectangle NW->SE, then try to go into the other quadrants, but only SE->NW is possible to do.

Pointer has a similar, separate behaviour, for which I'm opening an issue.

https://github.com/user-attachments/assets/8144dfb5-9aeb-4846-9089-964721fa5d4f

robertwidfen:
Good point - had fixed it in pointer yesterday and forgot it is here also 🙃

Now it should work.

BTW, the original square drawing will come back in next commit by Ctrl as a fallback to 1:1 aspect ratio for size zero.

PR #648, opened 2026-09-09
feat: add Ctrl modifier to keep resize aspect ratio

### LINKED ISSUE #633: Crop to specific aspect ratio/dimensions

robertwidfen:
Closes #633

See also #647

Brings old Shift (square drawing) back with Ctrl as a "side" effect,
for size (0, 0) - as when beginning a drag - we cannot determine the 
aspect ration so we use 1:1 as a fallback and thus create a square.


RobertMueller2:
lgtm

PR #654, opened 2026-09-11
fix: show center marker in pointer tool

### LINKED ISSUE #653: Pointer tool does not show center marker
### Version

0.23.0 dev

### Description

When using the Alt modifier (center resize) in the pointer tool it does not show the center marker when drawing a new shape.

robertwidfen:
Fixes #653

Also change the state vars across all tools to be the same name,
i.e. `editing` and not `finishing`.

Draw center marker always last, as it will not be visible on filled
shapes otherwise.

### DIFF
--- a/src/tools/blur.rs
+++ b/src/tools/blur.rs
@@ -127,10 +127,6 @@ impl Drawable for Blur {
             bounds,
         );
         if self.editing {
-            if self.centered {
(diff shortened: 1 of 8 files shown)


RobertMueller2:
lgtm, merging with one of the successor PRs ;)

PR #357, opened 2025-12-16
feat: Display image and crop dimensions in toolbar

prime-run:
This PR adds image and crop dimensions display to the toolbar for better visibility and fixed-size rendering.

- Show full image dimensions in toolbar (bottom right) when crop tool is inactive
- Display crop selection dimensions when crop tool is active  

NOTE: event flow from `CropTool → SketchBoard → App → StyleToolbar`

251216_11h22m24s_screenshot

-------

Update: [eab8720](https://github.com/Satty-org/Satty/pull/357/commits/eab87205bfbded43d6f82437c23f4368160b2a9e) Adds:
- Esc key cancels crop selection instead of killing process when crop tool is active.

Issue #649, opened 2026-09-09
Precise measurement of very small selections


Dupond:
### Description

Hello!

It would be great to have a dedicated measurement tool. Alternatively, the existing Crop tool could be improved so that very small selections can be measured reliably. The Crop tool already displays the dimensions of the current selection; however, its 2 px border is drawn centered on the selection boundary. As a result, part of the border overlaps the selected area.

This becomes problematic for very small selections. For example, with a 1×2 px selection, the border can completely obscure the pixels that are supposed to be measured.

A possible solution would be to:

* draw the crop border entirely outside the selected rectangle;
* use a 1 px, non-antialiased border;
* or otherwise make the selection boundary visually distinct without covering pixels inside the selection.

The important point is that a selection such as 1×2 px or 2×2 px should remain completely visible while its dimensions are displayed. Maybe a dedicated measurement tool would also be a great solution...

Thanks again for this great tool, and have a nice day!

robertwidfen:
For crop it soon will be fixed - the change is already on the way - I just now saw this issue here and that it will be related.

<img width="600" height="600" alt="Image" src="https://github.com/user-attachments/assets/a360c8d7-1f77-4e73-8fd5-b96c3859bae1" />

PR #655, opened 2026-09-11
feat: draw high-contrast helper elements

### LINKED ISSUE #194: Visibility of helper elements
With helper elements I mean anything that isn't actually drawn, e.g.
- the blur rectangle
- the recently added rectangle and center indicators for the ellipse tool

As they are mono color, It depends on the image whether these are visible. E.g. black ish rectangle on dark background isn't great.

I could see two different solutions to this:
- add contrast to each helper, so dark shapes are paired with bright outline or vice versa
- make the helper color configurable with its own button so it is an ad hoc decision for the user

robertwidfen:
See #194, #649, #357

Closes #194 - high-contrast helper elements
Closes #357 - crop usability improvements

Features:
- implements a two-color (out)line to ensure visibility
  on both dark and light backgrounds for pointer selection
  crop border and center marker.
- helper element sizes are now all zoom independent, i.e.
  stay the same when zooming in or out.
- set-out crop selection border to not cover cropped region
- set-out crop shadow to hide image completely
- increase opacity of crop selection when not editing it -
  this way it feels more like a applied crop.

### DIFF
--- a/src/tools/blur.rs
+++ b/src/tools/blur.rs
@@ -127,10 +127,6 @@ impl Drawable for Blur {
             bounds,
         );
         if self.editing {
-            if self.centered {
-                draw_center_marker(canvas, self.origin);
-            }
-
             // set style
             let mut color = Color::black();
             color.set_alphaf(0.6);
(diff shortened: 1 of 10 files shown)


RobertMueller2:
lgtm, merging with one of the successor PRs ;)

PR #657, opened 2026-09-11
fix: make selection border zoom independent

### LINKED ISSUE #650: Setout of the selection border should be zoom independent
### Description

When zooming deeply into the canvas it becomes too large.

It was intentionally added with #607 to make small annotations editable by avoiding overlap of handles and to not cover small lines, e.g. of rectangle.

Image

robertwidfen:
Closes #650

The handles now will sit on the bounds of the shape and
the selection border sits on the outer border of the
handles and thus still will not overlay small shapes.

Also when zoomed in hit test is zoom independent.


RobertMueller2:
lgtm, merging with one of the successor branches

Issue #339, opened 2025-11-13
Feature discussion: determine output canvas size


RobertMueller2:
Discussed in #110 and #336. Overlaps with #337 and #288.

Annotations can be (partially) outside of the image, there should be a way to save more canvas than just the image. 

- Should there be a silent default, e.g. use the outer bounds of all annoations, regardless of current visibility?
- Should there be a way to explicitly set outer bounds? Would this be an improved crop feature?
  - if so, should there be an actual applied crop on top of this?

robertwidfen:
In [gradia ](https://gradia.alexandervanhee.be/) one can add a background image and screenshot and all annotations sit on top of it.

I would not use a silent default, but allow crop to be larger - at the same time I have just added a "fix: limit crop selection to image bounds" which contradicts - maybe it should snap to image bounds ...

RobertMueller2:
I'm afraid if we allow to write a larger area, we need to control what is displayed in it, e.g. background colour, or indeed different image. I don't really want to open that can of worms.

That said, for dragging stuff, it might feel easier if you can use the whole area. So long it's clear that is not being in the output image, that's fine with me.

IIRC there used to be an issue with text exceeding the image or cropped area on screen, but the text was actually squeezed inside of the output image, causing a wysiwyg mismatch issue. Perhaps we could double check, if we don't have any of those wysiwyg issues anymore, I'm happy to just close this issue.

PR #658, opened 2026-09-11
feat: snap crop selection to image bounds

### LINKED ISSUE #649: Precise measurement of very small selections

robertwidfen:
Closes #649 - crop border not covering selection

See #339 (export larger area than image) why I do
not want to limit it to the image bounds for now.

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -407,8 +407,9 @@ impl FemtoVgAreaMut {
     // Hit-test all drawables and return all indices whose bounds contain `pos`, in order from topmost to bottommost.
     pub fn hit_test(&self, pos: Vec2D) -> Vec<usize> {
         let mut results = Vec::new();
+        let tolerance = crate::tools::HIT_BORDER_TOLERANCE / self.scale_factor;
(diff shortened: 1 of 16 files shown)


RobertMueller2 (approved):
lgtm, but see #666: diagonals might need a tad more tolerance. For now, I will merge this with one of the next successor commits.

PR #660, opened 2026-09-11
fix: avoid commit of micro drawables

robertwidfen:
They can be had to detect and grab - most tools already check for Vec2D::zero() or similar, but everything with a width/height below 1 or area below 1 (kind of duplicates width/height check) makes no sense.


RobertMueller2:
I think this should trigger a redraw somewhere if at all possible, otherwise it can leave stuff on the screen until next redraw.

PR #662, opened 2026-09-11
feat: export crop selection outside image

### LINKED ISSUE #339: Feature discussion: determine output canvas size

robertwidfen:
See also #633, #337 

Closes #339

The background will be transparent so people can put the
saved/copied image on top of other background in another
tool. Maybe in the future we can allow to move annotations
behind the image - but for now this is IMHO sufficient.

### DIFF
--- a/src/femtovg_area/imp.rs
+++ b/src/femtovg_area/imp.rs
@@ -699,6 +699,12 @@ impl FemtoVgAreaMut {
             preview.draw(canvas, font, bounds)?;
         }
 
+        // if not onscreen, flush and return early as the rest is not for export
+        if !onscreen {
+            canvas.flush();
+            return Ok(());
+        }
+
         // draw crop on top of everything but pointer tool selection overlay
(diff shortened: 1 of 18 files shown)


robertwidfen:
I keep it here for discussion and more testing ...

RobertMueller2:
This not a testing comment yet. Sounds like a decent solution for now.

This will inevitably lead to background colouring down the road, you read it here first ;)

robertwidfen:
yeah, but I was glad to realize that femtovg exports transparency which I feel as the most natural background and thus a good default.

PR #664, opened 2026-09-11
Feat/notify on config errors

### LINKED ISSUE #663: Config errors should *NOT* lead to abort of startup but be reported
### Description

When I start by Hyperland bindings and nothing happens it is odd.
Also if things get changed, deprecated or during testing it easier to just start.

robertwidfen:


RobertMueller2:
how about adding a logging component? Collect log messages in the background in a ring buffer (next to printing them to stderr), and have a button that is basically an error counter or something, once you click it, it opens a popup that displays the ring buffer. That way users could review any config errors even if they haven't started from terminal.

Notification only displays important stuff, like it is now.

PR #677, opened 2026-09-18
fix: point at `satty --help` instead of listing values by hand

### LINKED ISSUE #676: [Bug]: config.toml lists the wrong possible values for initial-tool
### Version

commit e6defa6 (current `main`)

### Description

The comment above `initial-tool` in `config.toml`, and the generated copy of it in `README.md`, does not match the values the option actually takes

The comment says:

```toml
# Select the tool on startup [possible values: pointer, crop, line, arrow, rectangle,
# text, marker, blur, brush, fringe-pixelate, fringe, fringe-pixelate, image]
```

`satty --help` derives its list from the `Tools` enum:

```
pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, pixelate, fringe-pixelate, fringe, highlight, brush, image
```

So the example config hides three tools that work — `ellipse`, `pixelate` and `highlight` — and prints `fringe-pixelate` twice, in the second case where `pixelate` belongs. Someone configuring Satty from `config.toml` alone has no way to learn that `initial-tool = "highlight"` is valid

The deprecated `action-on-enter` has the same problem: its comment names `save-to-clipboard, save-to-file`, while the option takes the whole `Action` enum — `save-to-clipboard, save-to-file, save-to-file-as, copy-filepath-to-clipboard, exit`, as `--help` shows

Both comments are hand-written copies of lists clap already derives from the enums, which is why they fall behind whenever a tool or an action is added

I would like to fix this by dropping the enumeration from both comments and pointing at `satty --help` instead:

```toml
# Select the tool on startup [possible values: see `satty --help`]
initial-tool = "brush"
```

That leaves a single place to maintain, and `make update-readme` carries it into `README.md`. The change is two comment lines plus the regenerated README, and I have it ready to open as a PR

If you would rather keep the values spelled out in the example config, that is fine too — say so and I will send the corrected lists instead

rokokol:
Closes #676

The comment above `initial-tool` in `config.toml` spelled the accepted values out by hand, and that copy had drifted from the `Tools` enum clap derives `--help` from: `ellipse`, `pixelate` and `highlight` were missing, while `fringe-pixelate` appeared twice, the second time in the position where `pixelate` belongs. The deprecated `action-on-enter` had drifted the same way, naming two of the five `Action` variants

Instead of correcting the copies, both comments now point at `satty --help`, which prints the values clap derives from the enums. That leaves one place to maintain, so the example config cannot fall behind a newly added tool again

`README.md` here is only the regenerated copy: the edit is in `config.toml`, and `make update-readme` carried it over. Nothing in the README was touched by hand, and the `satty --help` section it embeds is unchanged by this PR

No Rust code is involved, so `cargo fmt` and `cargo clippy` have nothing to say about the change

Tested with:
- `make update-readme`, which built `target/release/satty` and rewrote the README; the resulting diff is exactly the two comment lines in each file
- `satty --help`, which still prints `pointer, crop, line, arrow, rectangle, ellipse, text, marker, blur, pixelate, fringe-pixelate, fringe, highlight, brush, image` for `--initial-tool` — the list the comment now sends the reader to

If you would rather keep the values spelled out in the example config, say the word and I will replace this with the corrected lists instead


RobertMueller2:
Thanks!

